PR #1, opened 2016-10-10
Add AspNetCore meta package

BrennanConroy:
https://github.com/aspnet/Coherence-Signed/issues/378
@pranavkm @Eilon @muratg


NuGetPackageVerifier.json
```diff
@@ -0,0 +1,15 @@
+{
+    "adx": { // Packages written by the ADX team and that ship on NuGet.org
```
pranavkm: You can remove this node.

makefile.shade
```diff
@@ -0,0 +1,53 @@
...
+use-standard-lifecycle
+k-standard-goals
+
+var Configuration_Local = '${E("Configuration")}'
+default BASE_DIR_LOCAL='${Directory.GetCurrentDirectory()}'
+default TARGET_DIR_LOCAL='${Path.Combine(BASE_DIR_LOCAL, "artifacts")}'
+default BUILD_DIR_LOCAL='${Path.Combine(TARGET_DIR_LOCAL, "build")}'
+default SRC_PROJECT_GLOB_LOCAL="src/*/project.json"
+
+#build-compile target='compile' if='Directory.Exists("src")'
```
pranavkm: Why do we need this?

src/Microsoft.AspNetCore/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyConfiguration("")]
```
pranavkm: Clean this up - most of these things aren't needed.

src/Microsoft.AspNetCore/project.json
```diff
@@ -0,0 +1,27 @@
+﻿{
+  "version": "1.0.1-*",
```
pranavkm: Is this for the 1.0.0 branch? In which case, this would go in the 1.0.0 branch and you'd have to use the KoreBuild from one of the master branches of other repos (it pins versions)

src/Microsoft.AspNetCore/project.json
```diff
@@ -0,0 +1,27 @@
...
+    "Microsoft.AspNetCore.StaticFiles": "1.0.0",
+
+    "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.0",
+    "Microsoft.Extensions.Configuration.Json": "1.0.0",
+    "Microsoft.Extensions.Logging": "1.0.0",
+    "Microsoft.Extensions.Logging.Console": "1.0.0",
+    "Microsoft.Extensions.Options.ConfigurationExtensions": "1.0.0",
+    "Microsoft.Extensions.Configuration.FileExtensions": "1.0.0",
+
+    "Microsoft.AspNetCore.Routing": "1.0.1",
```
pranavkm: Weird ordering

src/Microsoft.AspNetCore/Properties/AssemblyInfo.cs
Eilon: Most not needed? There shouldn't be any .cs files at all, no?

src/Microsoft.AspNetCore/project.json
Eilon: This is how they were grouped in the spec. I think this might be fine for this package because it's... different. More of a logical grouping.

makefile.shade
BrennanConroy: To `DotNetPack("--no-build")`

src/Microsoft.AspNetCore/project.json
BrennanConroy: Read the attached issue, its 1.0.1

makefile.shade
pranavkm: Does the default makefile not work for this?

src/Microsoft.AspNetCore/project.json
pranavkm: 🙁  Why is Routing different from StaticFiles? Seems nicer to just alphabetize them like we do with every other repo.

NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
```
pranavkm: Could you copy the NuGet.config from https://github.com/aspnet/Common/blob/master/NuGet.config?

makefile.shade
BrennanConroy: It will make a nupkg with a .dll

pranavkm: In which case, can you simplify this? `AddAssemblyInfo` wouldn't apply so it could be reduced to `srcProject.ForEach(DotnetPack)`

makefile.shade
```diff
@@ -0,0 +1,53 @@
...
+            {
+              Console.WriteLine("Embedding commit hash in assembly");
+              var projectFolder = Path.GetDirectoryName(projectFile);
+              var commitHashAttribute = String.Format("[assembly: System.Reflection.AssemblyMetadata(\"CommitHash\", \"{0}\")]", commitHash);
+
+              var buildInfoFile = Path.Combine(projectFolder, "BuildInfo.generated.cs");
+              File.WriteAllText(buildInfoFile, commitHashAttribute);
+            }
+        }
+        DotnetPack(projectFile, BUILD_DIR_LOCAL, Configuration_Local, " --no-build");
```
pranavkm: Could you declare `BUILD_DIR` and `Configuration` inside this target? You wouldn't need to mess around with suffixing it in that case.

src/Microsoft.AspNetCore/Microsoft.AspNetCore.xproj
```diff
@@ -0,0 +1,21 @@
...
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
+
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>6f3d43f7-9546-4b41-af04-cf4708b62051</ProjectGuid>
+    <RootNamespace>Microsoft.AspNetCore</RootNamespace>
```
pranavkm: Remove

makefile.shade
```diff
@@ -0,0 +1,53 @@
+use import="Environment"
+
+var VERSION='0.1'
```
pranavkm: I don't think you need these variables.

src/Microsoft.AspNetCore/project.json
Eilon: I'm fine either way.

BrennanConroy:
Removed `makefile` because @davidfowl pointed out the `assemblyinfo.cs` file might be causing the `.dll` to be created, and he was right.

BrennanConroy:
Ping

PR #13, opened 2017-02-02
Upgrade to VS 2017

natemcmaster:
cref  https://github.com/aspnet/Universe/issues/469


makefile.shade
```diff
@@ -1,6 +0,0 @@
-var AUTHORS='Microsoft Open Technologies, Inc.'
-
-use-standard-lifecycle
-k-standard-goals
-
--Environment.SetEnvironmentVariable("KOREBUILD_ADD_ASSEMBLY_INFO", "0");
```
BrennanConroy: Where did this go :(

natemcmaster: It shouldn't matter. We don't include the binaries in this nupkg.

BrennanConroy:
:shipit:

PR #14, opened 2017-02-09
Set references to PrivateAssets=None

natemcmaster:
By default, pack sets `PrivateAssets="Build,Analyzers"`, which means the metapackage would not import targets from any of the packages listed here.

Changes the generated nuspec from:
```xml
<dependency id="Microsoft.AspNetCore.Diagnostics" version="1.2.0-preview1-23373" exclude="Build,Analyzers" />
```
to
```xml
<dependency id="Microsoft.AspNetCore.Diagnostics" version="1.2.0-preview1-23373" include="All" />
```

None of them do...yet...but we'll need this for MvcPrecompliation.

### DIFF
--- a/src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
+++ b/src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
@@ -10,17 +10,18 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


natemcmaster:
:bell:

BrennanConroy:
:shipit:

Issue #24, opened 2017-04-06
Add new WebHost static API to Microsoft.AspNetCore meta-package with opinionated defaults


DamianEdwards:
Add new static API `WebHost` to support creating opinionated instances of `IWebHost` and `IWebHostBuilder` which default the following over and above the usual defaults:
- Server to Kestrel
- Content root to working directory
- Other ideas  for defaults (in ascending order of craziness):
  - Log to console
  - Add cmd line and environment variable configuration providers, maybe even JSON for `appsettings.json` and `appsettings.[environment].json` too
  - Add the developer exception page middleware via an `IStartupFilter` by default when `IHostingEnvironment.IsDevelopment()`

There would be methods that allow building and starting an `IWebHost` with one call, and another that allows creating an `IWebHostBuilder` for more advanced scenarios, e.g. setting up configuration, logging, etc.
``` c#
namespace Microsoft.AspNetCore
{
    public static class WebHost
    {
        public static IWebHost Start(RequestDelegate app) { }

        public static IWebHost Start(string url, RequestDelegate app) { }

        public static IWebHost StartWith(Action<IApplicationBuilder> app) { }

        public static IWebHost StartWith(string url, Action<IApplicationBuilder> app) { }
        
        public static IWebHostBuilder CreateBuilder() { }
    }
}
```

They would be used like so:
``` c#
namespace MyApp
{
    public static class Program
    {
        public static void Main(string[] args)
        {
            // Hello, World!
            var host1 = WebHost.Start(context => context.Response.WriteAsync("Hello, World!"));
            host1.WaitForShutdown();

            // Hello, World! + middleware
            var host2 = WebHost.StartWith(app =>
            {
                var env = app.ApplicationServices.GetService<IHostingEnvironment>();
                if (env.IsDevelopment())
                {
                    app.UseDeveloperExceptionPage()
                }
                
                app.Run(context => context.Response.WriteAsync("Hello, World!"))
            );
            host2.WaitForShutdown();

            // Getting a builder to do something more complex, the templates would do this
            var host3 = WebHost.CreateBuilder()
                .UseLogging(logging => logging
                    .AddConsole()
                    .AddDebug()
                )
                .UseStartup<Startup>()
                .Build();
            host3.Run();
        }
   }
}
```

## Some slightly more audacious ideas
Add overloads that allow easy creation of an app that uses a single routed middleware:
``` c#
public static IWebHost Start(Action<IRouteBuilder> routes) { }
[...]

PR #26, opened 2017-04-07
[DESIGN] Adds new WebHost API for creating IWebHost & IWebHostBuilder with def…

DamianEdwards:
Adds the API detailed in #24


src/Microsoft.AspNetCore/WebHostStartupFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+        public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
+        {
+            return app =>
+            {
+                var env = app.ApplicationServices.GetService<IHostingEnvironment>();
+                if (env.IsDevelopment())
+                {
+                    app.UseDeveloperExceptionPage();
+                }
+            };
```
Tratcher: Need to call next

DamianEdwards: Doh

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -0,0 +1,195 @@
...
+        private static IWebHost StartWith(string url, Action<IServiceCollection> configureServices, Action<IApplicationBuilder> app)
+        {
+            var builder = CreateDefaultBuilder();
+
+            if (!string.IsNullOrEmpty(url))
+            {
+                builder.UseUrls(url);
+            }
+
+            if (configureServices != null)
```
davidfowl: Why do we need configure services?

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            return builder;
+        }
+
+        private static Assembly GetApplicationAssembly()
+        {
+            try
+            {
+                // Copied from https://github.com/aspnet/PlatformAbstractions/blob/dev/src/Microsoft.Extensions.PlatformAbstractions/ApplicationEnvironment.cs#L49
+                var getEntryAssemblyMethod =
```
davidfowl: We shouldn't be using this. Target ns 1.5 and use the public API

DamianEdwards: 👍

src/Microsoft.AspNetCore/WebHost.cs
DamianEdwards: Yep, routing

src/Microsoft.AspNetCore/WebHost.cs
DamianEdwards: Should we cross-target to ns1.5 and only put an assembly in that lib, or the entire meta-package?

DamianEdwards:
@davidfowl could you give a yay or nay on this please so I can hand this off to dev for completion, thanks.

src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <!-- set PrivateAssets=None to ensure that all assets, including Build and Analyzer, are included in the nuspec -->
+    <PackageReference Include="Microsoft.AspNetCore" Version="$(AspNetCoreVersion)" PrivateAssets="None" />
```
davidfowl: Why wasn't this included before?

src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
```diff
@@ -19,6 +23,8 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="$(AspNetCoreVersion)" PrivateAssets="None" />
     <PackageReference Include="Microsoft.Extensions.Configuration.FileExtensions" Version="$(AspNetCoreVersion)" PrivateAssets="None" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="$(AspNetCoreVersion)" PrivateAssets="None" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="$(AspNetCoreVersion)" PrivateAssets="None" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="$(AspNetCoreVersion)" PrivateAssets="None" />
```
davidfowl: So this is now in the minimal package?

DamianEdwards: Yes I had to add them to achieve the desired defaults.

src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
DamianEdwards: Not sure, guessing it was a judgement based on it having no assets of its own previously.

JunTaoLuo: As I mentioned previously I only included the files marked as ship,include in the packages.csv in Coherence: https://github.com/aspnet/Coherence/blob/dev/packages/packages.csv#L3. Microsoft.AspNetCore was marked as exclude so I did not include it in the M.A.All package. Given that the M.A.All package is a superset of the M.A metapackage I'm not sure if this is required.

JunTaoLuo: Actually, we should include it if there is now code we need to ship in the M.A metapackage.

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -0,0 +1,179 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore
+{
+    /// <summary>
+    /// Provides convenience methods for creating instances of <see cref="IWebHost"/> and <see cref="IWebHostBuilder"/> with pre-configured defaults.
+    /// </summary>
+    public static class WebHost
```
Tratcher: We already have a WebHost class. This will at least cause naming conflicts.
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L24-L26.

DamianEdwards: Oh then @davidfowl lied to me 😀
The other one is in an internal namespace so we can rename if need be. This is unlikely to ever cause naming conflicts in an app as ppl don't import our internal namespaces. Are there places in our code it causes an issue? I'd really like this type name for this class.

davidfowl: We can make that one internal.

Tratcher: Could also use the katana name WebApp, it already has many of the same methods.

davidfowl: I don't think we should use any names from katana.

src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
```diff
@@ -3,10 +3,12 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <IncludeBuildOutput>false</IncludeBuildOutput>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <IncludeBuildOutput>true</IncludeBuildOutput>
```
Tratcher: this is the default, correct? can you remove it?

src/Microsoft.AspNetCore/WebHost.cs
Tratcher: @JunTaoLuo you're likely to break Hosting with ambiguous type reference errors.

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        /// <summary>
+        /// Initializes and starts a new <see cref="IWebHost"/> with pre-configured defaults.
+        /// See <see cref="CreateDefaultBuilder()"/> for details.
+        /// </summary>
+        /// <param name="app">A delegate that handles requests to the application.</param>
+        /// <returns>A started <see cref="IWebHost"/> that hosts the application.</returns>
+        public static IWebHost Start(RequestDelegate app) =>
+            Start(null, app);
```
Tratcher: named parameters

JunTaoLuo:
Merging now to unblock https://github.com/aspnet/MetaPackages/issues/30. Will add tests in a later PR.

PR #38, opened 2017-04-21
Static API second attempt

JunTaoLuo:
The only difference from the change yesterday is https://github.com/aspnet/MetaPackages/commit/47d90ba7232145299226131b77965c235327bf5a
which changes 
`Assembly.GetEntryAssembly()` to `Assembly.Load(new AssemblyName(env.ApplicationName))`

This allows us to continue targeting netstandard1.3 instead of netstandard1.5. We want to keep targeting netstandard1.3 since we want the metapackage to work for net46 apps without having to cross-compile net46 for the metapackage.

cc @davidfowl @cesarbs @mikeharder


JunTaoLuo:
FYI only, merged in https://github.com/aspnet/MetaPackages/commit/d9661ed70dc4f6d8347d956f787484d8d59ef3e4

PR #40, opened 2017-04-22
WebHost static API updates

JunTaoLuo:
#24 following up to more issues.

- Specify ApplicationName when wrapping RequestDelegate or RouteBuilder
- Add debug logger by default

Still need to add tests.

### DIFF
--- a/samples/SampleApp/SampleApp.csproj
+++ b/samples/SampleApp/SampleApp.csproj
@@ -12,6 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


JunTaoLuo:
Split out the tests to #46 since I suspect it will have more feedback. This change can go in now.

samples/SampleApp/SampleApp.csproj
```diff
@@ -8,10 +8,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Folder Include="wwwroot\" />
+    <Content Include="wwwroot\htmlpage.html" />
```
Tratcher: old publish include?

PR #41, opened 2017-04-23
Use new Hosting APIs

JunTaoLuo:
https://github.com/aspnet/Hosting/pull/1035

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -162,9 +162,9 @@ namespace Microsoft.AspNetCore
                         config.AddCommandLine(args);
                     }
                 })
-                .ConfigureLogging(logging =>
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -162,9 +162,9 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
                         config.AddCommandLine(args);
                     }
                 })
-                .ConfigureLogging(logging =>
+                .ConfigureLogging((context, logging) =>
                 {
-                    logging.AddConsole();
+                    logging.AddConsole(context.Configuration);
```
JunTaoLuo: Slight behaviour difference from before but I think it's reasonable to pass the configuration to the console logger by default.

JunTaoLuo:
Closing in favour of https://github.com/aspnet/MetaPackages/pull/54

Issue #39, opened 2017-04-21
Using AspNetCore.All package just as a metapackage on net46 does not work


kichalla:
Repro:
- Clone MusicStore
- change directory to `samples\MusicStore`
- Change the `samples\MusicStore.csproj` tfms to only have `net46`
- dotnet restore
- dotnet build

Expected:
build succeeds

Actual:
build fails
```
Models\ManageViewModels.cs(2,29): error CS0234: The type or namespace name 'DataAnnotations' does not exist in the namespace 'System.ComponentModel' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
Models\ManageViewModels.cs(3,17): error CS0234: The type or namespace name 'AspNetCore' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
Models\ManageViewModels.cs(4,17): error CS0234: The type or namespace name 'AspNetCore' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
Models\ManageViewModels.cs(5,17): error CS0234: The type or namespace name 'AspNetCore' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
Models\MusicStoreContext.cs(1,17): error CS0234: The type or namespace name 'AspNetCore' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
Models\MusicStoreContext.cs(2,17): error CS0234: The type or namespace name 'EntityFrameworkCore' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
Models\Order.cs(2,29): error CS0234: The type or namespace name 'DataAnnotations' does not exist in the namespace 'System.ComponentModel' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
Models\Order.cs(3,17): error CS0234: The type or namespace name 'AspNetCore' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?) [C:\github\musicstore\samples\MusicStore\MusicStore.csproj]
[...]

Issue #21, opened 2017-04-05
Add missing packages to M.A.All


JunTaoLuo:
Tracking issues regarding inclusion of our packages in the metapackage:

- [ ] https://github.com/aspnet/AzureIntegration/issues/20. This doesn't block the inclusion of the affected package in the cache but we will be cross-gen'ing the 1.0.0 and 2.0.0 versions of Hosting and a few other packages until the dependency is updated
- [ ] https://github.com/Azure/azure-sdk-for-net/issues/3003. Blocking inclusion of Microsoft.Extensions.Configuration.AzureKeyVault
- [x] https://github.com/aspnet/DataProtection/pull/217. Blocking inclusion of Microsoft.AspNetCore.DataProtection.AzureStorage
- [x] Coherence needs to pass after AzureIntegration renames. Blocking inclusion of AzureIntegration packages.

DamianEdwards:
It also needs all the toolish dependency packages, e.g. Microsoft.AspNetCore.Mvc.Razor.ViewCompilation, Microsoft.EntityFrameworkCore.Design, etc.

JunTaoLuo:
Parsing from the [packages.csv](https://github.com/aspnet/Coherence/blob/dev/packages/packages.csv) (which should be updated since we plan on making that the source of truth) https://github.com/aspnet/MetaPackages/issues/19

To be included:
===============
```
Microsoft.AspNetCore.Mvc.Razor.ViewCompilation,ship,exclude
Microsoft.CodeAnalysis.Razor.Workspaces,ship,exclude
Microsoft.EntityFrameworkCore.Design,ship,exclude
Microsoft.EntityFrameworkCore.Relational.Design,ship,exclude
Microsoft.EntityFrameworkCore.Sqlite.Design,ship,exclude
Microsoft.EntityFrameworkCore.SqlServer.Design,ship,exclude
Microsoft.Extensions.Caching.Redis,ship,exclude
Microsoft.VisualStudio.Web.BrowserLink,ship,exclude
```
Not included:
=============

Pre-release:
------------
```
Microsoft.AspNetCore.Buffering,noship,exclude
Microsoft.AspNetCore.DataProtection.Redis,noship,exclude
Microsoft.AspNetCore.Diagnostics.Elm,noship,exclude
Microsoft.AspNetCore.Diagnostics.Identity.Service,noship,exclude
Microsoft.AspNetCore.Identity.Service.Core,noship,exclude
[...]

PR #43, opened 2017-04-24
Add more package to package cache.

JunTaoLuo:
Addresses #21 #39 

I'm not able to add Microsoft.CodeAnalysis.Razor.Workspaces however since it has a PCL dependency: Microsoft.Composition 1.0.27

cc @Eilon @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
+++ b/src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <IncludeBuildOutput>false</IncludeBuildOutput>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
     <PackageTags>aspnetcore</PackageTags>
     <Description>Microsoft.AspNetCore.All</Description>
     <EnableApiCheck>false</EnableApiCheck>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <IncludeBuildOutput>false</IncludeBuildOutput>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
JunTaoLuo: I had to bump this to netcoreapp2.0 since M.A.Diagnostics.Identity.Service targets net46/netcoreapp2.0. Any reason why this package is not targeting netstandard1.x? @javiercn

javiercn: We need Certificate Signing APIs that are only available in .NET Core 2.0

javiercn:
LGTM (The identity service part)

src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
Eilon: I *think* this is fine, but I'd like @DamianEdwards to confirm this is ok. This means the metapackage can't be used on .NET Framework, but I think for 2.0 that's fine.

JunTaoLuo:
Clarified issues with @DamianEdwards offline in person.

Issue #30, opened 2017-04-17
Bind Kestrel options to config by default


danroth27:
_From @danroth27 on April 17, 2017 20:5_

Since config is in DI now by default we should support binding the Kestrel options to config by default (for `UseKestrel()`).

```json
  "Kestrel": {
    "Endpoints": [ // Order matters!
      {
        "Address": "localhost",
        "Port": "44316",
        "Certificate": { // Implies HTTPS
          "Subject": "danroth27.com", // Pick the cert that matches the subject name with the latest expiry
          "StoreName": "My",
          "StoreLocation": "CurrentUser"
        }
      },
      {
        "Address": "localhost",
        "Port": "8080"
      },
      {
        "Address": "localhost",
        "Port": "44317",
        "Certificate": {
          "Path": "./mysslcert.pfx",
          "Password": "abc!123" // In user secrets: Kestrel:Endpoints:2:Certificate:Password
        }
      }
    ]
  }
```

_Copied from original issue: aspnet/KestrelHttpServer#1703_

danroth27:
@glennc @Tratcher @DamianEdwards @davidfowl

danroth27:
_From @muratg on April 17, 2017 20:53_

:eyes:

danroth27:
Or you can name your endpoints:

```json
  "Kestrel": {
    "Endpoints": {
      "CertFromStore": {
        "Address": "localhost",
        "Port": "44316",
        "Certificate": { // Implies HTTPS
          "Subject": "danroth27.com", // Pick the cert that matches the subject name with the latest expiry
          "StoreName": "My",
          "StoreLocation": "CurrentUser"
        }
      },
      "LocalhostWithPort": {
        "Address": "localhost",
        "Port": "8080"
      },
      "CertFromFile": {
        "Address": "localhost",
        "Port": "44317",
        "Certificate": {
          "Path": "./mysslcert.pfx",
          "Password": "abc!123" // In user secrets: Kestrel:Endpoints:CertFromFile:Certificate:Password
        }
      }
    }
  }

danroth27:
We decided to do this in the new static `WebHost` API.

danroth27:
@muratg Please assign for preview1

muratg:
@JunTaoLuo could you take this? Please sync up with @Tratcher.

danroth27:
A couple of updates:
- We will need to be able to configure certificates for multiple purposes: SSL, token signing, etc.
[...]

PR #44, opened 2017-04-24
Bind Kestrel options to config by default (#30)

cesarbs:
#30


src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -148,6 +150,7 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
 
                     if (env.IsDevelopment())
                     {
+                        Console.WriteLine(env.ApplicationName);
```
davidfowl: I assume this is just left over.

src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
```diff
@@ -11,7 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-   <PackageReference Include="@(MetaPackagePackageReference)" />
+    <PackageReference Include="@(MetaPackagePackageReference)" />
```
davidfowl: Undo this change?

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore
+{
+    /// <summary>
+    /// Binds Kestrel configuration.
+    /// </summary>
+    public class KestrelServerOptionsSetup : IConfigureOptions<KestrelServerOptions>
+    {
+        private IServiceProvider _services;
```
davidfowl: This should be `IConfiguration`, not `IServiceProvider`

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        /// Creates a new instance of <see cref="KestrelServerOptionsSetup"/>.
+        /// </summary>
+        /// <param name="services">An <seealso cref="IServiceProvider"/> instance.</param>
+        public KestrelServerOptionsSetup(IServiceProvider services)
```
davidfowl: See comment above. Same thing.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// </summary>
+        /// <param name="certificateConfiguration">An <seealso cref="IConfiguration"/> with information about a certificate source.</param>
+        /// <param name="password">The certificate password, in case it's being loaded from a file.</param>
+        /// <returns>The loaded certificates.</returns>
+        public static X509Certificate2 Load(IConfiguration certificateConfiguration, string password = null)
+        {
+            var sourceKind = certificateConfiguration.GetValue<string>("Source");
+
+            CertificateSource certificateSource;
+            switch (sourceKind)
```
davidfowl: Is this supposed to be case sensitive?

cesarbs: cc @javiercn

src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
cesarbs: Spacing was wrong.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// Loads all certificates specified in an <seealso cref="IConfiguration"/>.
+        /// </summary>
+        /// <param name="configurationRoot">The root <seealso cref="IConfiguration"/>.</param>
+        /// <returns>
+        /// A dictionary mapping certificate names to loaded certificates.
+        /// </returns>
+        public static Dictionary<string, X509Certificate2> LoadAll(IConfiguration configurationRoot)
+        {
+            var certificates = configurationRoot.GetSection("Certificates");
+            var loadedCertificates = new Dictionary<string, X509Certificate2>();
```
davidfowl: This looks like it could be a single LINQ statement.

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,86 @@
...
+                BindEndPoint(options, endPoint, certificates);
+            }
+        }
+
+        private void BindEndPoint(
+            KestrelServerOptions options,
+            IConfigurationSection endPoint,
+            Dictionary<string, X509Certificate2> certificates)
+        {
+            options.Listen(IPAddress.Parse(endPoint.GetValue<string>("Address")), endPoint.GetValue<int>("Port"), listenOptions =>
```
davidfowl: This needs to be more failure tolerant. TryParse with a decent exception message

samples/AppSettings/Program.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using Microsoft.AspNetCore;
+using Microsoft.AspNetCore.Http;
+
+namespace AppSettings
+{
+    public class Program
+    {
+        public static void Main(string[] args)
```
Tratcher: Why not use the existing sample?

src/Microsoft.AspNetCore/CertificateLoader.cs
Tratcher: No, avoid treating config as case sensitive.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        private class CertificateStoreSource : CertificateSource
+        {
+            public string Subject { get; set; }
+            public string StoreName { get; set; }
+            public string StoreLocation { get; set; }
+
+            public override X509Certificate2 Load()
+            {
+                if (!Enum.TryParse(StoreLocation, true, out StoreLocation storeLocation))
```
Tratcher: named params

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    throw new InvalidOperationException($"Invalid store location: {StoreLocation}");
+                }
+
+                using (var store = new X509Store(StoreName, storeLocation))
+                {
+                    store.Open(OpenFlags.ReadOnly);
+                    var foundCertificate = store.Certificates.Find(X509FindType.FindBySubjectName, Subject, validOnly: false)
```
Tratcher: Is there an EKU check anywhere?

samples/AppSettings/appsettings.json
```diff
@@ -0,0 +1,61 @@
+{
+  "Kestrel": {
+    "EndPoints": {
+      "Http": {
+        "Address": "127.0.0.1",
+        "Port": 8081
+      },
+      "HttpV6": {
+        "Address": "[::1]",
```
Tratcher: Brackets not required

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            options.Listen(address, port, listenOptions =>
+            {
+                var certificateName = endPoint.GetValue<string>("Certificate");
+
+                X509Certificate2 endPointCertificate = null;
+                if (certificateName != null)
+                {
+                    endPointCertificate = certificates[certificateName];
```
Tratcher: KeyNotFoundException? You can give a better error.

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,101 @@
...
+                }
+                else
+                {
+                    var certificate = endPoint.GetSection("Certificate");
+
+                    if (certificate.GetChildren().Any())
+                    {
+                        var password = _configurationRoot[$"Kestrel:EndPoints:{endPoint.Key}:Certificate:Password"];
```
Tratcher: is this not already part of the certificate config section? even if it's coming from a different config source it should get merged.

samples/AppSettings/Program.cs
cesarbs: Settings would apply to all the scenarios there. I want to keep things separate for clarity.

samples/SampleApp/SampleApp.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
-  <Import Project="..\..\build\common.props" />
+  <Import Project="..\..\build\dependencies.props" />
```
davidfowl: ?

cesarbs: @natemcmaster pointed out we don't normally import `common.props` here, only `dependencies.props`. This was causing issues with user secrets because `common.props` has `GenerateUserSecretsAttribute` set to false.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+                }
+
+                return certificate;
+            }
+
+            private X509Certificate2 TryLoad(X509KeyStorageFlags flags, out Exception exception)
+            {
+                try
+                {
+                    var loadedCertificate = new X509Certificate2(Path, _password);
```
javiercn: var loadedCertificate = new X509Certificate2(Path, _password, flags);

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            public string Path { get; set; }
+
+            public override X509Certificate2 Load()
+            {
+                var certificate = TryLoad(X509KeyStorageFlags.DefaultKeySet, out var error)
+                    ?? TryLoad(X509KeyStorageFlags.UserKeySet, out error)
+    #if NETCOREAPP2_0
+                    ?? TryLoad(X509KeyStorageFlags.EphemeralKeySet, out error)
```
cesarbs: @blowdart Is this the proper order to attempt to load the cert? cc @javiercn

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,174 @@
...
+            private X509Certificate2 TryLoad(X509KeyStorageFlags flags, out Exception exception)
+            {
+                try
+                {
+                    var certificate = new X509Certificate2(Path, _password, flags);
+
+                    if (!HasServerAuthEKU(certificate))
+                    {
+                        throw new InvalidOperationException(
+                            $"The certificate file at '{Path}' does not contain an Enhanced Key Usage field with the OID 1.3.6.1.5.5.7.3.1 (Server Authentication)");
```
cesarbs: @blowdart Is the wording correct?

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,174 @@
...
+                        .FirstOrDefault();
+
+    #if NET46
+                    store.Close();
+    #endif
+
+                    if (foundCertificate == null)
+                    {
+                        throw new InvalidOperationException(
+                            $"No valid certificate found for {Subject} in store {StoreName} under {StoreLocation}. Check that the certificate contains the Enhanced Key Usage field with the OID 1.3.6.1.5.5.7.3.1 (Server Authentication)");
```
cesarbs: @blowdart Is the wording correct?

cesarbs:
@danroth27 Should the EKU check be mandatory as I've done it here? Or should we add a flag to optionally disable it?

Also: should we actually move the check to Kestrel?

src/Microsoft.AspNetCore/CertificateLoader.cs
cesarbs: Added.

danroth27:
Yeah, if we're going to check the EKUs at all it should be done lower in the stack. Let's remove this check from here.

cesarbs:
Removed EKU check. Needs further discussion.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,140 @@
...
+
+                using (var store = new X509Store(StoreName, storeLocation))
+                {
+                    store.Open(OpenFlags.ReadOnly);
+                    var foundCertificate = store.Certificates.Find(X509FindType.FindBySubjectName, Subject, validOnly: false)
+                        .OfType<X509Certificate2>()
+                        .OrderByDescending(certificate => certificate.NotAfter)
+                        .FirstOrDefault();
+
+    #if NET46
```
davidfowl: No needed.

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,103 @@
...
+            var portValue = endPoint.GetValue<string>("Port");
+
+            IPAddress address;
+            if (!IPAddress.TryParse(addressValue, out address))
+            {
+                throw new InvalidOperationException($"Invalid IP address: {addressValue}");
+            }
+
+            int port;
+            if (!int.TryParse(portValue, out port))
```
Tratcher: culture, style

cesarbs:
@halter73 Added you as a reviewer.

src/Microsoft.AspNetCore/CertificateLoader.cs
blowdart: Why validOnly: false? Do we want people to be able to bind to expired certs?

src/Microsoft.AspNetCore/CertificateLoader.cs
blowdart: What are you trying to do here? @bartonjs because he understands Ephemeral :)

blowdart:
@danroth27 It's not a question of IF we perform an EKU check. We should/must. So if it's removed from here, then where is the corresponding issue in kestrel to perform the check there?

danroth27:
@blowdart Filed https://github.com/aspnet/KestrelHttpServer/issues/1764 to track validating the cert EKU

src/Microsoft.AspNetCore/CertificateLoader.cs
bartonjs: I read this as:

* Load perphemeral, as MachineKeySet as the PFX says so, or as UserKeySet if it has no opinion on the matter.
* Load perphemeral as UserKeySet (in case the key was MachineKeySet and the current process does not have write access to the machine key store).
* Load ephemeral (in case the current process has no persisted key write access at all, such as low rights with no user profile)

src/Microsoft.AspNetCore/CertificateLoader.cs
bartonjs: * FindBySubjectName does strstr. If you want exact match you'll need to post-process it.
* store.Certificates returns unique X509Certificate2 objects, and Find returns unique X509Certificate2 objects.  If you would like to reduce Finalizations you should dispose all of the members of store.Certificates (which you need to save as a local, since reading it twice just makes the problem worse).

Similary, if you cascade call Find you should dispose all the previous values.

cesarbs:
📢

src/Microsoft.AspNetCore/CertificateLoader.cs
cesarbs: @javiercn Since this was originally your code, what are your thoughts here?

javiercn: @blowdart self-signed certificates. I would change that to HostingEnvironment.IsDevelopment() if its available

javiercn: Another option is to just load the certificates and filter using Linq, or have an extra configuration value per certificate that indicates if its ok to load invalid (so that you can again load self-signed certificates for development)

Yet another option is to make the check true and force the user to put the certificate (in case its self-signed) in the trusted certificate store for .NET. (Which varies across platforms)

cesarbs: Talked to @danroth27 and we decided to go with a config key (`AllowInvalid`).

cesarbs:
@bartonjs Can you review the changes I made to loading the certs from the store? Disposing & validating subject further.

bartonjs:
Cert/store loading looks good to me.  You might consider plumbing your EKU check to a `.Where` so that you don't pick a new-improper cert over a slightly-older-but-valid cert.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,167 @@
...
+                    X509Certificate2Collection foundCertificates = null;
+
+                    try
+                    {
+                        store.Open(OpenFlags.ReadOnly);
+                        storeCertificates = store.Certificates;
+                        foundCertificates = storeCertificates.Find(X509FindType.FindBySubjectName, Subject, validOnly: !AllowInvalid);
+                        var foundCertificate = foundCertificates
+                            .OfType<X509Certificate2>()
+                            .Where(certificate => certificate.Subject.Equals(Subject, StringComparison.Ordinal))
```
danroth27: Why do we need this where clause when we already did a find by subject name?

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,103 @@
...
+                    if (!certificates.TryGetValue(certificateName, out endPointCertificate))
+                    {
+                        throw new InvalidOperationException($"No certificate named {certificateName} found in configuration");
+                    }
+                }
+                else
+                {
+                    var certificate = endPoint.GetSection("Certificate");
+
+                    if (certificate.GetChildren().Any())
```
danroth27: What happens if the Certificate property isn't there at all on the endpoint? Would this nullref?

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (!int.TryParse(portValue, out port))
+            {
+                throw new InvalidOperationException($"Invalid port: {portValue}");
+            }
+
+            options.Listen(address, port, listenOptions =>
+            {
+                var certificateName = endPoint.GetValue<string>("Certificate");
+
+                X509Certificate2 endPointCertificate = null;
```
danroth27: It would be nice if this logic for dealing with a property that could be a certificate name or an inline certificate was exposed as a helper method on the certificate loader so that we can use it in other places where we want the same behavior (like configuring a signing certificate).

src/Microsoft.AspNetCore/CertificateLoader.cs
cesarbs: @bartonjs pointed out that `FindSubjectByName` does a substring search, so we need this extra check to get an exact string match.

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
cesarbs: No, you get an empty section.

src/Microsoft.AspNetCore/CertificateLoader.cs
bartonjs: Yeah, then he learned he was both right and wrong.  Silly @bartonjs.  You want FindByDistinguishedSubjectName, that leaves all of the RDN keys in place, and matches on String.Equals.  So it does this exact check for you, and you can then drop the Where.

(FindBySubjectName drops all the tags out, for... reasons?, and also doesn't reverse the fields.  So where you want to match "CN=localhost, OU=fun times, C=ZZ" it matches "ZZ, fun times, localhost")

FindByDistinguishedName matches IgnoreCase for some reason, so if you want to be case-respecting you might want to leave this here after all.

cesarbs: When I use `FindBySubjectDistinguishedName`, I have to change my settings to `CN=localhost` to load my certificate (and I guess that's what you meant by the other option dropping the tags).

@danroth27 Do we want this or should we keep the previous behavior?

src/Microsoft.AspNetCore/CertificateLoader.cs
blowdart: "perphemeral"? 🤣 

Is there an advantage in trying "perphemeral" first, if ephermeral is there? There's no real need to persist the private key anyhere.

cesarbs:
@muratg Adding you as a reviewer since this will require preview1 approval.

src/Microsoft.AspNetCore/CertificateLoader.cs
danroth27: We discussed this offline and we agreed we want to use the distinguished subject name (i.e. CN=localhost)

muratg:
@cesarbs will you send a PR to the preview1 branch as well?  Also, when is the user-secrets work be done?

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
danroth27: Some examples of how we would like to reuse the cert loading logic: https://github.com/aspnet/MetaPackages/issues/30#issuecomment-297527522

cesarbs:
@muratg I've rebased this to `rel/2.0.0-preview1`. Once it's merged I'll merge it to `dev`.

cesarbs:
@muratg User secrets working depends on https://github.com/dotnet/cli/pull/6356.

cesarbs:
@JunTaoLuo @Tratcher Can you guys review the functional tests I've just added?

src/Microsoft.AspNetCore/CertificateLoader.cs
cesarbs: I'll put the call with `EphemeralKeySet` first then, with a fallback to `UserKeySet` if that fails. Does that sound right?

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -0,0 +1,173 @@
...
+                var certificate = TryLoad(X509KeyStorageFlags.DefaultKeySet, out var error)
+                    ?? TryLoad(X509KeyStorageFlags.UserKeySet, out error)
+    #if NETCOREAPP2_0
+                    ?? TryLoad(X509KeyStorageFlags.EphemeralKeySet, out error)
+    #endif
+                    ;
+
+                if (error != null)
+                {
+                    throw error;
```
Tratcher: Rethrowing this exception looses the original stack trace. You should wrap it or use ExceptionDispatchInfo to capture and rethrow it.

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,103 @@
...
+        private void BindEndPoint(
+            KestrelServerOptions options,
+            IConfigurationSection endPoint,
+            Dictionary<string, X509Certificate2> certificates)
+        {
+            var addressValue = endPoint.GetValue<string>("Address");
+            var portValue = endPoint.GetValue<string>("Port");
+
+            IPAddress address;
+            if (!IPAddress.TryParse(addressValue, out address))
```
Tratcher: Wait, I can't bind to * or localhost? That's a non-starter.

Tratcher: https://github.com/aspnet/MetaPackages/issues/73

PR #51, opened 2017-04-25
Set TargetManifestFiles in PropertyGroup

### LINKED ISSUE #50: Publish output not trimmed when referencing Microsoft.AspNetCore.All
After the target was added in https://github.com/aspnet/MetaPackages/issues/20. It is not currently trimming the publish output.

JunTaoLuo:
Fixes #50 

It's too late to set the `TargetManifestFiles` property inside a target that runs before publish. The dotnet sdk publish targets resolve the file list [here](https://github.com/dotnet/sdk/blob/39102ec214514363c6af4cd9d599a35aceb8c745/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Publish.targets#L32-L34). Since we don't rely on any other targets or properties to set `TargetManifestFiles`, the proposal in thiss PR sets the property in a property group before the donet sdk publish targets. Note that in this approach we no longer produce an error if the resolution of the manifest rid fails and will fail silently instead.

I tested on a local dotnet new console app with a package reference to M.A.All package and the publish output is trimmed.

### DIFF
--- a/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
+++ b/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
@@ -1,37 +1,37 @@
 <Project>
-
   <PropertyGroup>
     <PublishWithAspNetCoreTargetManifest Condition="'$(PublishWithAspNetCoreTargetManifest)'==''">true</PublishWithAspNetCoreTargetManifest>
+    <ManifestRuntimeIdentifier Condition="'$(PublishWithAspNetCoreTargetManifest)'=='true' and '$(PublishableProject)'=='true'" >$(AspNetCoreTargetManifestRuntimeIdentifier)</ManifestRuntimeIdentifier>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(PublishWithAspNetCoreTargetManifest)'=='true' and '$(PublishableProject)'=='true' and '$(ManifestRuntimeIdentifier)'==''">
+    <ManifestRuntimeIdentifier
+      Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))'
+        and '$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture)'=='X64'">win7-x64</ManifestRuntimeIdentifier>
+    <ManifestRuntimeIdentifier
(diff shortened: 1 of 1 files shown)


davidfowl:
That's why I originally asked why we didn't just set a property.

src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
```diff
@@ -31,18 +15,9 @@ Append the default ASP.NET Core runtime package store manifest during publish
         Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))'">osx.10.12-x64</ManifestRuntimeIdentifier>
       <ManifestRuntimeIdentifier
         Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))'">linux-x64</ManifestRuntimeIdentifier>
-    </PropertyGroup>
-
-    <Error
```
pranavkm: What happens if we leave the target around (but leave the resolution globally)? Can we get an error message in that case?

src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
```diff
@@ -1,26 +1,10 @@
 <Project>
-
   <PropertyGroup>
     <PublishWithAspNetCoreTargetManifest Condition="'$(PublishWithAspNetCoreTargetManifest)'==''">true</PublishWithAspNetCoreTargetManifest>
+    <ManifestRuntimeIdentifier>$(AspNetCoreTargetManifestRuntimeIdentifier)</ManifestRuntimeIdentifier>
```
pranavkm: `<ManifestRuntimeIdentifier Condition="'$(PublishWithAspNetCoreTargetManifest)'=='true' and '$(PublishableProject)'=='true'">$(AspNetCoreTargetManifestRuntimeIdentifier)</ManifestRuntimeIdentifier>`

src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
JunTaoLuo: Sure, I wasn't sure whether that was valuable though.

Issue #58, opened 2017-04-26
WebHost.CreateDefaultBuilder should enable DI scope validation when in Development


DamianEdwards:
Instead of doing it in templates, we're going to do it in `WebHost.CreateDefaultBuilder`

PR #61, opened 2017-04-26
Validate scopes in Development environment in CreateDefaultBuilder

pakrym:
Fixes: https://github.com/aspnet/MetaPackages/issues/58
/cc @DamianEdwards


pakrym:
@JunTaoLuo filed https://github.com/aspnet/MetaPackages/issues/72 will do for preview2

PR #62, opened 2017-04-26
Suppress failures to add user secrets

### LINKED ISSUE #60: Exception when UserIdSecrets isn't turned on in the project file
I created a 1.1.1 template and upgraded to 2.0 and used the new API the got this error:

image

It's pretty awful that this throws like that by default. Is there any way to avoid this exception and have it noop (like the other config providers do) if the application isn't configured for user secrets?

/cc @natemcmaster

natemcmaster:
Two things can go wrong:

1. Assembly.Load failures if env.ApplicationName does not correspond to a valid assembly
1. No UserSecretsIdAttribute on the assembly

Resolves #60

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -159,10 +159,17 @@ namespace Microsoft.AspNetCore
 
                     if (env.IsDevelopment())
                     {
-                        var appAssembly = Assembly.Load(new AssemblyName(env.ApplicationName));
-                        if (appAssembly != null)
+                        try
                         {
-                            config.AddUserSecrets(appAssembly);
+                            var appAssembly = Assembly.Load(new AssemblyName(env.ApplicationName));
+                            if (appAssembly != null)
+                            {
+                                config.AddUserSecrets(appAssembly);
+                            }
+                        }
+                        catch
+                        {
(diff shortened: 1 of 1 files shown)


davidfowl:
> Assembly.Load failures if env.ApplicationName does not correspond to a valid assembly

Super rare. Just let it explode. Other parts of the stack try to `Load(env.ApplicationName)` already so don't bother.

> No UserSecretsIdAttribute on the assembly

I'd rather add detection using reflection than by catching exceptions. No first chance exceptions in ASP.NET Core! 😄

natemcmaster:
How about instead we add an optional setting to the .AddUserSecrets call?

davidfowl:
> How about instead we add an optional setting to the .AddUserSecrets call?

That was my first choice. Make it like the other providers.

DamianEdwards:
A bool to suppress errors?

natemcmaster:
I was thinking of adding `.AddUserSecrets(Assembly assembly, bool optional)` and `.AddUserSecrets<T>(bool optional)`

DamianEdwards:
Yup

natemcmaster:
K, let's to that. See https://github.com/aspnet/Configuration/pull/654.

I'll open a PR to use the optional = true setting when it's it.

PR #64, opened 2017-04-26
AddUserSecrets is optional

### LINKED ISSUE #60: Exception when UserIdSecrets isn't turned on in the project file

natemcmaster:
Depends on https://github.com/aspnet/Configuration/pull/654

Resolves #60


DamianEdwards:
:shipit:

PR #66, opened 2017-04-26
Produce tar.gz file for osx and linux package store

### LINKED ISSUE #48: Produce Linux package store as a tarball instead of zip
Our docker images don't have `unzip` installed. If we produce a tarball instead, then we can use `tar` to extract the package cache.

cc @JunTaoLuo

natemcmaster:
Resolves #48 

Required for https://github.com/aspnet/aspnet-docker/issues/215


JunTaoLuo:
gah this is going to conflict with #67

muratg:
@natemcmaster @JunTaoLuo Do we need to sit down and talk about how the folder structure look like across different archives and installers?

natemcmaster:
I trust @JunTaoLuo knows these details. I'm not familiar with folder structure expectations. This change is simply about converting a .zip to a .tar.gz so Docker install is possible.

JunTaoLuo:
I don't see why the folder structure would need to be different on different archives? Why do we have this concern? Also, this PR needs to be rebased but looks correct

PR #70, opened 2017-04-27
Package Fallback Folder archive generation

anurse:
1. Ignore the specific packages we're using (in Archive.AspNetCore.All), its like this because of current build issues in ASP.NET Core, so I can test the process despite our packages being in a broken state

2. All of `dotnet-archive` and `Microsoft.DotNet.Archive` are copied as-is from https://github.com/dotnet/cli/tree/master/src

The main logic is in repo.targets.

/cc @muratg @Eilon


JunTaoLuo:
post-hoc :shipit:

PR #78, opened 2017-04-27
use actual set of packages

anurse:
This adds the proper set of packages to the Package Archive list


NuGet.config
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
+    <add key="msbuild" value="https://dotnet.myget.org/F/msbuild/api/v3/index.json" />
```
pranavkm: @anurse - what's this feed for?

anurse: Earlier work when it looked like I needed to pull specific msbuild packages. I don't think it's needed anymore.

PR #79, opened 2017-04-28
Hack a no timestamp version of the runtime store and deps file

JunTaoLuo:
This looks so ugly. Especially when I have to shell out to ps/bash for search and replace since I didn't find a good way of doing it in MSBuild.


JunTaoLuo:
Merged in https://github.com/aspnet/MetaPackages/commit/485cf19ab6364fd415a6cb2a903b64da335ab68d

PR #82, opened 2017-04-28
Use RuntimeFrameworkVersion instead for directory name

pakrym:


src/Microsoft.AspNetCore.RuntimeStore/Microsoft.AspNetCore.RuntimeStore.csproj
```diff
@@ -17,6 +17,6 @@
         </DepsFiles>
       </ItemGroup>
 
-      <Copy SourceFiles="%(DepsFiles.FullPath)" DestinationFiles="$(DepsOutputPath)\%(DepsFiles.PackageName)\Microsoft.NETCore.App\2.0.0\deps.json" />
+      <Copy SourceFiles="%(DepsFiles.FullPath)" DestinationFiles="$(DepsOutputPath)\%(DepsFiles.PackageName)\Microsoft.NETCore.App\$(RuntimeFrameworkVersion)\deps.json" />
```
JunTaoLuo: @glennc also found that we need to add `shared` into this path. Also the file name needs to be <something>.deps.json instead of just deps.json.

so `DestinationFiles="$(DepsOutputPath)\%(DepsFiles.PackageName)shared\\Microsoft.NETCore.App\$(RuntimeFrameworkVersion)\%(DepsFiles.PackageName).deps.json"`

src/Microsoft.AspNetCore.RuntimeStore/Microsoft.AspNetCore.RuntimeStore.csproj
```diff
@@ -7,16 +7,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-   <PackageReference Include="@(RuntimeStorePackageReference)" />
+   <PackageReference Include="@(RuntimeStorePackageReference)"  Version="$(AspNetCoreVersion)" PrivateAssets="None" />
```
JunTaoLuo: why the version and private assets attributes?

src/Microsoft.AspNetCore.HostingStartup.Template/Program.cs
```diff
@@ -0,0 +1,4 @@
+public class Program
```
JunTaoLuo: hate to put a copyright header on this but we probably should.

build/repo.targets
```diff
@@ -55,7 +58,9 @@
     <RemoveDir Directories="$(PackageCacheOutputPath)" />
     <RemoveDir Directories="$(WorkingDirectory)" />
     <Exec Command="dotnet store --manifest $(MetaPackageFile) --configuration Release --framework netcoreapp2.0 --runtime $(RID) --output $(PackageCacheOutputPath) --framework-version 2.0.0-* --working-dir $(WorkingDirectory)" />
-    <MsBuild Projects="$(MetaPackageFile)" Targets="CollectDeps" Properties="DepsOutputPath=$(DepsOutputPath)"/>
+
+    <!--- MSBuild caches things if you run inproc so have to use Exec -->
+    <Exec Command="dotnet msbuild /t:Restore;Rebuild;CollectDeps $(HostingStartupTemplateFile) /p:DepsOutputPath=$(DepsOutputPath);HostingStartupPackageName=%(HostingStartupPackageReference.Identity);HostingStartupPackageVersion=%(Version)"/>
```
JunTaoLuo: 👍

src/TrimDeps/TrimDeps.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
```
JunTaoLuo: put the version in dependencies.props.

src/Microsoft.AspNetCore.HostingStartup.Template/Microsoft.AspNetCore.HostingStartup.Template.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="$(HostingStartupPackageName)" Version="$(HostingStartupPackageVersion)" PrivateAssets="None" />
+  </ItemGroup>
+
+  <Target Name="CollectDeps">
+    <PropertyGroup>
+      <DestinationDepsFile>$(DepsOutputPath)\$(HostingStartupPackageName)\shared\Microsoft.NETCore.App\$(RuntimeFrameworkVersion)\deps.json</DestinationDepsFile>
```
JunTaoLuo: must have some kind of file name in front of deps.json. Could be as simple as app.deps.json

src/TrimDeps/Program.cs
```diff
@@ -0,0 +1,58 @@
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+public class Program
+{
+    public static void Main(string[] args)
+    {
+        ChangeEntryPointLibraryName(args[0], null);
```
JunTaoLuo: I'm a little confused what is trying to be achieved here. From the code below it looks like you are only removing the target library which is the first entry under targets/tfm and libraries. In that case why bother with the second parameter? Also is it always guaranteed that the first entry is what we want to trim? Is it better to be logging for the specific package  *HostingStartup.Template I presume?

pakrym: Sorry, should have added WIP to commit message, I copied this from the code I wrote to build shared runtime I'm going to trim it today.

tools/TrimDeps/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+        using (var file = File.OpenText(depsFile))
+        using (JsonTextReader reader = new JsonTextReader(file))
+        {
+            deps = JObject.ReadFrom(reader);
+        }
+
+        string version = null;
+        foreach (JProperty target in deps["targets"])
+        {
+            var targetLibrary = target.Value.Children<JProperty>().FirstOrDefault();
```
JunTaoLuo: Is this logic too fragile? Are we guaranteed that the first entry is always the one we want to remove? Should we not check the name of the assembly or something?

pakrym: https://github.com/dotnet/core-setup/blob/6887ab556bc8302390782711dcdd75b75e769cf5/src/test/build/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs#L149 we use it to build shared runtime, I think it's good enough.

JunTaoLuo: :shipit:

Issue #69, opened 2017-04-27
Update certificate loader to support diferent patterns for loading collections of certificates


danroth27:
Update the certificate loader to support the following scenarios:

```json
{
    "Authentication": {
        "Identity": {
            "SigningCertificates": { // Multiple inline
                "MySigningCert1": {
                    "Source": "File",
                    "Path": "cert1.pfx",
                    "Password": "*****"
                },
                "MySigningCert1": {
                    "Source": "File",
                    "Path": "cert2.pfx",
                    "Password": "*****"
                }
            }
        }
    }
}
```
```json
{
    "Authentication": {
        "Identity": {
            "SigningCertificates": { // Single inline
                "Source": "File",
                "Path": "cert1.pfx",
                "Password": "*****"
            }
        }
    }
}
```
```json
{
    "Certificates": {
        "MySigningCert1": {
            "Source": "File",
            "Path": "cert1.pfx",
            "Password": "*****"
        },
    },
    "Authentication": {
        "Identity": {
            "SigningCertificates": "MySigningCert1" // Single referenced
        }
    }
}
```
```json
{
    "Certificates": {
        "MySigningCert1": {
            "Source": "File",
            "Path": "cert1.pfx",
            "Password": "*****"
        },
        "MySigningCert2": {
            "Source": "File",
            "Path": "cert2.pfx",
            "Password": "*****"
        }
    },
    "Authentication": {
        "Identity": {
            "SigningCertificates": "MySigningCert1 MySigningCert2" // Multiple referenced, space separated
        }
    }
}
```

cesarbs:
preview1? cc @muratg

PR #85, opened 2017-04-28
Support more certificate loading scenarios (#69)

cesarbs:
#69


src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -65,23 +121,18 @@ private abstract class CertificateSource
...
-    #if NETCOREAPP2_0
-                    ?? TryLoad(X509KeyStorageFlags.EphemeralKeySet, out error)
-    #endif
-                    ;
+                Exception error;
+                var certificate =
+#if NETCOREAPP2_0
+                    TryLoad(X509KeyStorageFlags.EphemeralKeySet, out error) ??
+#endif
+                    TryLoad(X509KeyStorageFlags.UserKeySet, out error);
```
cesarbs: @blowdart @bartonjs Does this look right?

Actually, unless I'm missing something, the `#if` here doesn't make sense since we're targeting `netstandard1.3` only. Should I only leave it as `UserKeySet`?

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -12,23 +12,90 @@ namespace Microsoft.AspNetCore
...
+        }
+
+        /// <summary>
+        /// Loads one or more certificates based on the information found in a configuration section.
+        /// </summary>
+        /// <param name="certificateConfiguration">A configuration section containing either a string value referencing certificates
+        /// by name, or one or more inline certificate specifications.
+        /// </param>
+        /// <returns>One or more loaded certificates.</returns>
+        public IEnumerable<X509Certificate2> Load(IConfigurationSection certificateConfiguration)
```
cesarbs: @danroth27 Do you want to keep the keys when loading from e.g.

```json
"MyCert" : {
  "Source": "File",
  "Path": "mycert.pfx"
},
"MyCert2": {
  "Source": "Store",
  "StoreName": "My",
  "StoreLocation": "CurrentUser"
}
```

?

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -12,23 +12,90 @@ namespace Microsoft.AspNetCore
...
     /// A helper class to load certificates from files and certificate stores based on <seealso cref="IConfiguration"/> data.
     /// </summary>
-    public static class CertificateLoader
+    public class CertificateLoader
     {
+        private readonly IConfiguration _certificatesConfiguration;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="KestrelServerOptionsSetup"/>.
+        /// </summary>
```
muratg: "Wrong!" :)

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -12,23 +12,90 @@ namespace Microsoft.AspNetCore
...
+        public CertificateLoader()
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="KestrelServerOptionsSetup"/> that can load certificates by name.
+        /// </summary>
```
muratg: Ditto

src/Microsoft.AspNetCore/CertificateLoader.cs
danroth27: By "keys" do you mean the certificate name used in config? Yeah, I think it would be good to provide those.

cesarbs:
📢

danroth27:
We shouldn't throw if we look for certificates in store and don't find any. Instead we should throw where those certificates are being used (ex identity/kestrel).

src/Microsoft.AspNetCore/CertificateLoader.cs
danroth27: But not required.

src/Microsoft.AspNetCore/CertificateLoader.cs
danroth27: @Eilon @DamianEdwards @muratg @davidfowl 
We should prefer ephemeral if it is available, which is only available in netcoreapp2.0. I believe that means this package (Microsoft.AspNetCore) needs to target netcoreapp2.0, not netstandard1.3.

muratg:
@cesarbs, @DamianEdwards thinks we should multi-target `netstandard1.3` and `netcoreapp2.0` so that we keep supporting desktop CLR apps.

src/Microsoft.AspNetCore/CertificateLoader.cs
cesarbs: When I put ephemeral first, I can't run the sample app. It loads the certificate but this happens on every request:

```
fail: Microsoft.AspNetCore.Server.Kestrel[0]
      Uncaught exception from the OnConnectionAsync method of an IConnectionAdapter.
System.ComponentModel.Win32Exception (0x80004005): No credentials are available in the security package
   at System.Net.SSPIWrapper.AcquireCredentialsHandle(SSPIInterface secModule, String package, CredentialUse intent, SCHANNEL_CRED scc)
   at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(CredentialUse credUsage, SCHANNEL_CRED secureCredential)
   at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(X509Certificate certificate, SslProtocols protocols, EncryptionPolicy policy, Boolean isServer)
   at System.Net.Security.SecureChannel.AcquireServerCredentials(Byte[]& thumbPrint)
   at System.Net.Security.SecureChannel.GenerateToken(Byte[] input, Int32 offset, Int32 count, Byte[]& output)
   at System.Net.Security.SecureChannel.NextMessage(Byte[] incoming, Int32 offset, Int32 count)
   at System.Net.Security.SslState.StartSendBlob(Byte[] incoming, Int32 count, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.ProcessReceivedBlob(Byte[] buffer, Int32 count, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.StartReceiveBlob(Byte[] buffer, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.ForceAuthentication(Boolean receiveFirst, Byte[] buffer, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.ProcessAuthentication(LazyAsyncResult lazyResult)
   at System.Net.Security.SslStream.BeginAuthenticateAsServer(X509Certificate serverCertificate, Boolean clientCertificateRequired, SslProtocols enabledSslProtocols, Boolean checkCertificateRevocation, AsyncCallback asyncCallback, Object asyncState)
   at System.Net.Security.SslStream.<>c__DisplayClass35_0.<AuthenticateAsServerAsync>b__0(AsyncCallback callback, Object state)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl(Func`3 beginMethod, Func`2 endFunction, Action`1 endAction, Object state, TaskCreationOptions creationOptions)
   at System.Threading.Tasks.TaskFactory.FromAsync(Func`3 beginMethod, Action`1 endMethod, Object state)
   at System.Net.Security.SslStream.AuthenticateAsServerAsync(X509Certificate serverCertificate, Boolean clientCertificateRequired, SslProtocols enabledSslProtocols, Boolean checkCertificateRevocation)
   at Microsoft.AspNetCore.Server.Kestrel.Https.HttpsConnectionAdapter.<OnConnectionAsync>d__7.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.FrameConnection.<ApplyConnectionAdaptersAsync>d__32.MoveNext()
```

So I'm leaving the order as it is for now.

PR #87, opened 2017-04-29
Include tar.gz files in Build.RS nupkg

JunTaoLuo:
After https://github.com/aspnet/Coherence-Signed/pull/478 we need to include .tar.gz files in the Build.RS nupkg. 

Note that this package is used mainly for testing with JitBench on x-plat. cc @rynowak This is not blocking installer work since they obtain the stores from the drop share.

cc @muratg to decide whether this should go into rel/2.0.0-preview1 or only dev.

### DIFF
--- a/build/Build.RS.nuspec
+++ b/build/Build.RS.nuspec
@@ -8,5 +8,6 @@
   </metadata>
   <files>
     <file src="*.zip" target="" />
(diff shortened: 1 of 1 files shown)


muratg:
cc @mikeharder

PR #91, opened 2017-05-02
Minimal restore when building deps files

JunTaoLuo:
This should fix the CI package cache build flakiness.

cc @kichalla @danroth27


JunTaoLuo:
FYI only, merging to try to reduce build flakiness.

PR #93, opened 2017-05-02
Use win7-x64/x86 rids for runtime store

JunTaoLuo:
We should be building the store witn win7-x64 and win7-x86 rids instead of win-x64 and win-x86 rids.

We should also verify that linux-x64 and osx-x64 are sufficient for the xplat runtime stores. It would be a dramatic change if we build one for each possible platform.

cc @DamianEdwards @davidfowl


muratg:
@davidfowl can you please take a look?

JunTaoLuo:
We discussed this offline after lunch. Pushing this change in for the build now. Will speak to @davidfowl offline to verify a few more things.

Issue #83, opened 2017-04-28
Provide a nice error message when the HTTPS certificate is not configured for the current environment


danroth27:
Currently if you have a kestrel endpoint configured with a named certificate that isn't configured for the current environment you get an exception telling you that the certificate is missing. We should handle this exception and provide a reasonable error message to the user.

Current behavior:

```
~\Desktop\templates2\mvc-individual> $env:ASPNETCORE_ENVIRONMENT=""
~\Desktop\templates2\mvc-individual> dotnet run

Unhandled Exception: System.InvalidOperationException: No certificate named HTTPS found in configuration
   at Microsoft.AspNetCore.KestrelServerOptionsSetup.<>c__DisplayClass4_0.<BindEndPoint>b__0(ListenOptions listenOptions)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.Listen(IPEndPoint endPoint, Action`1 configure)
   at Microsoft.AspNetCore.KestrelServerOptionsSetup.BindEndPoint(KestrelServerOptions options, IConfigurationSection endPoint, Dictionary`2 certificates)
   at Microsoft.AspNetCore.KestrelServerOptionsSetup.BindConfiguration(KestrelServerOptions options)
   at Microsoft.Extensions.Options.LegacyOptionsCache`1.CreateOptions()
   at System.Threading.LazyInitializer.EnsureInitializedCore[T](T& target, Boolean& initialized, Object& syncLock, Func`1 valueFactory)
   at Microsoft.Extensions.Options.LegacyOptionsCache`1.get_Value()
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer.CreateServiceContext(IOptions`1 options, ILoggerFactory loggerFactory)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer..ctor(IOptions`1 options, ITransportFactory transportFactory, ILoggerFactory loggerFactory)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
[...]

PR #94, opened 2017-05-02
Show fwlink on HTTPS certificate errors (#83)

cesarbs:
#83


src/Microsoft.AspNetCore/CertificateConfigurationException.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore
+{
+    internal class CertificateConfigurationException : Exception
+    {
+        public CertificateConfigurationException(string message)
+            : base(message + ". For information on configuring HTTPS see https://go.microsoft.com/fwlink/?linkid=848054.")
```
muratg: Why not `$"{message}. For infor...."`?

Eilon: Why is it adding a period to the previous message at all?

Eilon: O I C. All the previous exception messages should be complete sentences (wherever possible) and end with periods... Though some of them I can see why they don't...

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -99,7 +99,7 @@ private X509Certificate2 LoadSingle(string certificateName)
 
             if (!certificateConfiguration.Exists())
             {
-                throw new InvalidOperationException($"No certificate named {certificateName} found in configuration");
+                throw new CertificateConfigurationException($"No certificate named {certificateName} found in configuration for the current environment");
```
Eilon: This sentence needs to end with a period.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -119,7 +119,7 @@ private X509Certificate2 LoadSingle(IConfigurationSection certificateConfigurati
                     certificateSource = new CertificateStoreSource(_certificateStoreLoader);
                     break;
                 default:
-                    throw new InvalidOperationException($"Invalid certificate source kind: {sourceKind}");
+                    throw new CertificateConfigurationException($"Invalid certificate source kind: {sourceKind}");
```
Eilon: Change this to be a sentence: `Invalid certificate source kind '{sourceKind}'.`

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -163,7 +163,7 @@ public override X509Certificate2 Load()
 
                 if (error != null)
                 {
-                    throw error;
+                    throw new CertificateConfigurationException($"Unable to load certificate from file '{Path}': {error.Message}");
```
Eilon: `Unable to load certificate from file '{Path}'. Error details: '{error.Message}'.`

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -203,7 +203,7 @@ public override X509Certificate2 Load()
             {
                 if (!Enum.TryParse(StoreLocation, ignoreCase: true, result: out StoreLocation storeLocation))
                 {
-                    throw new InvalidOperationException($"Invalid store location: {StoreLocation}");
+                    throw new CertificateConfigurationException($"Invalid certificate store location: {StoreLocation}");
```
Eilon: `The certificate store location '{StoreLocation}' is invalid.`

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -63,7 +63,7 @@ private void BindConfiguration(KestrelServerOptions options)
 
                     if (certificate == null)
                     {
-                        throw new InvalidOperationException($"Unable to load certificate for endpoint '{endPoint.Key}'");
+                        throw new CertificateConfigurationException($"Unable to load certificate for endpoint '{endPoint.Key}'");
```
Eilon: End with period.

src/Microsoft.AspNetCore/CertificateConfigurationException.cs
Eilon: Actually I suggested fixes for all the exception messages below. Then remove the leading period on this one here.

danroth27:
The CertificateLoader will be used for other scenarios than loading the HTTPS certificate. We're using it to load signing certificates as well. So I don't think the information about how to enable HTTPS is being added in the right context. I.e. you don't need instructions about setting up HTTPS if the configuration for your signing certificate is invalid. I think the HTTPS information needs to be added when setting up HTTPS with Kestrel.

Also, we should log a warning if the CertificateLoader try to find a certificate with a particular subject name in a store but doesn't find any.

src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -99,7 +125,7 @@ private X509Certificate2 LoadSingle(string certificateName)
 
             if (!certificateConfiguration.Exists())
             {
-                throw new InvalidOperationException($"No certificate named {certificateName} found in configuration");
+                throw new KeyNotFoundException($"No certificate named '{certificateName}' found in configuration for the current environment.");
```
cesarbs: @Tratcher Is it ok to throw this exception type here?

src/Microsoft.AspNetCore/CertificateConfigurationException.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore
+{
+    internal class CertificateConfigurationException : Exception
```
cesarbs: @Tratcher Is it ok to keep this internal?

test/Microsoft.AspNetCore.FunctionalTests/CertificateLoaderTests.cs
```diff
@@ -988,12 +1013,78 @@ public void ReturnsNull_MultipleCertificatesInline_FileAndStore_NotFoundInStore(
...
+                null,
+                loggerFactory.Object,
+                Mock.Of<ICertificateFileLoader>(),
+                Mock.Of<ICertificateStoreLoader>());
+
+            var loadedCertificates = certificateLoader.Load(configuration.GetSection("TestConfig:Certificates"));
+            Assert.Equal(0, loadedCertificates.Count());
+            Assert.Single(logger.LogMessages, logMessage =>
+                logMessage.LogLevel == LogLevel.Warning &&
+                logMessage.Message == "Unable to find a matching certificate for subject 'localhost' in store 'My' in 'CurrentUser'.");
```
cesarbs: @Eilon Is the language alright? Should log messages also end with a period?

src/Microsoft.AspNetCore/CertificateLoader.cs
Tratcher: 👍

src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
```diff
@@ -27,7 +31,7 @@ public void Configure(KestrelServerOptions options)
 
         private void BindConfiguration(KestrelServerOptions options)
         {
-            var certificateLoader = new CertificateLoader(_configurationRoot.GetSection("Certificates"));
+            var certificateLoader = new CertificateLoader(_configurationRoot.GetSection("Certificates"), _loggerFactory);
```
cesarbs: @danroth27 I think we should put `CertificateLoader` in DI.

@davidfowl What do you think?

src/Microsoft.AspNetCore/CertificateConfigurationException.cs
Tratcher: :-1:

Issue #96, opened 2017-05-03
WebHostFunctionalTests.BindsKestrelHttpsEndPointFromConfiguration fails on OSX


mikeharder:
```
[Microsoft.AspNetCore.Tests.WebHostFunctionalTests.BindsKestrelHttpsEndPointFromConfiguration_ReferencedCertificateFile]
System.PlatformNotSupportedException : This platform does not support loading with EphemeralKeySet. Remove the flag to allow keys to be temporarily created on disk.

[Microsoft.AspNetCore.Tests.WebHostFunctionalTests.BindsKestrelHttpsEndPointFromConfiguration_InlineCertificateFile]
System.PlatformNotSupportedException : This platform does not support loading with EphemeralKeySet. Remove the flag to allow keys to be temporarily created on disk.
```

http://aspnetci/viewLog.html?buildId=214187&buildTypeId=XPlat_Mac_MacUniverseCoreCLR&tab=buildLog&filter=err#_focus=358135

PR #100, opened 2017-05-03
Fix HTTPS functional tests on macOS (#96)

cesarbs:
#96

HTTPS tests were using a self-signed certificate that had something weird about it that didn't work on Mac. Now I'm just using the other self-signed certificate in the repo and that works.

`HttpClientSlim` was copied from Kestrel because `HttpClient` doesn't work on Mac with self-signed certs after switching to Apple's own security libs. cc @stephentoub @bartonjs

I should move `HttpClientSlim` to https://github.com/aspnet/Testing but for now this unblocks the CI.


cesarbs:
@mikeharder Oops. Done.

mikeharder:
Please file an issue to move `HttpClientSlim` to a central location to eliminate duplicate code.

Issue #117, opened 2017-05-17
Add Microsoft.AspNet.Configuration targeting default configuration schema


HaoK:
•	Create a new Microsoft.AspNetCore.Configuration that will contain all of our default config schema/setup which lives only in the meta package.
•	It will provide a single services.ConfigureAspNetDefaults() [naming TBD] which will add all of the default config conventions, none of the existing AddXyz() methods will use IConfiguration (and we should potentially remove any overloads that exist today)
•	Implementation detail: a ConfigureAspNet that implements IConfigureOptions for all things we want to configure by default
•	For preview 2, this will consist of whatever is being configured in the templates today (Authentication, IdentityService, Kestrel)
•	Longer term: Have intellense/schema for the important config settings.
•	The structure of the config will be nested: example using today’s indidual Auth template
``` 
  "Microsoft": {
    "Logging": {
      "IncludeScopes": false,
      "Debug": {
        "LogLevel": {
          "Default": "Warning"
        }
      },
      "Console": {
        "LogLevel": {
          "Default": "Warning"
        }
      }
    },
    "AspNetCore": {
      "Authentication": {
        "IdentityService": {
          "ClientId": "C06ACE17-3212-454B-84AB-14DDB2FC58E0",
          "TokenRedirectUrn": "urn:self:aspnet:identity:integrated"
        }
      },
      "Hosting": {
        "Kestrel": { // https://github.com/aspnet/MetaPackages/blob/dev/src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
          "Endpoints": {
            "Localhost": {
              "Address": "127.0.0.1",
              "Port": "5000"
            },
            "LocalhostWithHttps": {
              "Address": "127.0.0.1",
              "Port": "44333",
              "Certificate": {
                "Source": "Store",
                "StoreLocation": "LocalMachine",
                "StoreName": "My",
                "Subject": "CN=localhost"
              }
            }
          }
        }
```

cwe1ss:
Does it really have to be *that* nested? This looks very XMLy 😄 

The "Microsoft" and "AspNetCore" containers seem to be unnecessary because they probably won't have properties on their own. Would be nicer if we could just use "Logging", "Authentication", "Hosting" (or maybe even "Kestrel") as top-level entries. People will be smart enough to use different titles for their own areas. 

It would also make it easier/more readable for flat-item providers -
[...]

PR #118, opened 2017-05-17
Add ConfigureAspNetCoreDefaults

HaoK:
Initial prototype of https://github.com/aspnet/MetaPackages/issues/117

- New ConfigureAspNetCoreDefaults() that will add all of the IConfigureOptions for the default sections
- Currently: Kestral/OIDC/Google/Facebook/Twitter/MSA/JwtBearer/Identity
- Moved kestrel config under Microsoft:AspNetCore:Hosting


Eilon (changes requested):
cc @davidfowl @DamianEdwards 

We need to think a bit more about which dependencies go where. We can't add too many deps to the mini-metapackage. But, if we add this to the All-metapackage, then it can't be used from .NET Framework apps/templates.

HaoK:
Updated using new options infrastructure.  Should the kestrel setup move to Kestrel instead of living here like the rest of the default setup?

HaoK:
@JunTaoLuo do you own MetaPackage?  Who should review this change?

HaoK:
Merged without the kestrel changes which would break templates without the corresponding config update

PR #119, opened 2017-05-19
Exclude build assets from sqlitepclraw and dependencies

JunTaoLuo:
Per https://github.com/NuGet/Home/issues/5233, we need to mark everything that transitively depends on sqlitepclraw as PrivateAssets="Build". I took a look a the EF packages and it seems like they don't have any build assets them selves so this should be safe. I have verified that this workaround works and the sqlite native binaries no longer shows up in the publish output. Will need to update buildtools with the new third-party dependencies.

There is discussion with regards to whether the changes in https://github.com/dotnet/sdk/pull/1133 will be changed and the net461 PackageTargetFallback removed so this may no longer be relevant. However, in the meantime, we can use this workaround to trim the publish output.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -164,16 +164,19 @@
     <FullMetaPackagePackageReference Include="Microsoft.AspNetCore.WebSockets" />
     <FullMetaPackagePackageReference Include="Microsoft.AspNetCore.WebUtilities" />
     <FullMetaPackagePackageReference Include="Microsoft.CodeAnalysis.Razor" />
-    <FullMetaPackagePackageReference Include="Microsoft.Data.Sqlite" />
+    <FullMetaPackagePackageReference Include="Microsoft.Data.Sqlite" PrivateAssets="Build" />
     <FullMetaPackagePackageReference Include="Microsoft.Data.Sqlite.Core" />
-     <!-- workaround for https://github.com/dotnet/sdk/issues/1166 and https://github.com/ericsink/SQLitePCL.raw/issues/157 -->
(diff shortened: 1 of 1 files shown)


Eilon:
This is... awful... do we have a bug to undo this in the future? I'm only ok with this if it's a workaround, which would mean there would need to be a commitment from the appropriate team that the underlying issue is getting fixed.

cc @davidfowl

davidfowl:
This was supposed to be temporary for preview1, I don't understand why we haven't made progress since then.

JunTaoLuo:
Yes this is meant to be the corrected version of the workaround for preview 1 which didn't work.  Again, it may not be needed pending discussion on the net461 implicit package target fallback. I'll forward the email chain regarding the discussion.

natemcmaster:
> This was supposed to be temporary for preview1, I don't understand why we haven't made progress since then.

We need the SDK team to do something about this https://github.com/dotnet/sdk/issues/1166

JunTaoLuo:
Seems like the discussion in the dotnet/sdk issue is suggesting that we actually need to make this change since ATF will still bring in the net35 target files. So it's likely that we need to take this workaround for preview2 to properly trim.

natemcmaster:
:shipit:, though it's regrettable that this is necessary.

JunTaoLuo:
In the email discussion it seems that the dotnet/sdk issue actually will fix this but it's not likely to be corrected for preview2. I will merge this workaround for now and have an issue to track undoing it once we receive the fix.

PR #120, opened 2017-05-19
Add support for a default SSL certificate in development

javiercn:
When the ssl certificate specified is 'localhost' and
no certificate has been explicitly specified in the certificates
section, fallback to a certificate with 'CN=localhost' (ASP.NET Core SSL
developer certificate) in the current user store and if not found,
an ssl certificate with 'CN=localhost' on the current machine
(IIS Express developer certificate)

### DIFF
--- a/src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
+++ b/src/Microsoft.AspNetCore/KestrelServerOptionsSetup.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


blowdart:
This needs to go hand in hand with my requested changes for @cesarbs to enhance certificate selection first to prefer localhost certificates which have a SAN entry for localhost, then if no valid ones are present, drop back to subject name.

cesarbs:
The code looks fine but ideally it should be somewhere where it can be easily testable i.e. you should be able to write a test with a mock store loader and verify that it's trying to load the expected things from the expected places.

PR #121, opened 2017-05-19
React to dotnet store cli arguments update

JunTaoLuo:
This will fix the coherence-packagecache builds @ryanbrandenburg


ryanbrandenburg:
:shipit: if it fixes the build, but why did that option disappear? Did we know it was going to happen and just miss a spot or is this a surprise? It's fine to fix the build but I want to understand the cause, both technical and organizational.

JunTaoLuo:
Yea it was known, they removed it a while ago since it didn't make sense for the tool to take in a configuration. We probably just ingested the build that has the updates.

PR #123, opened 2017-05-24
Push unsigned all package with manifests

JunTaoLuo:
This allows us to use the aspnetcore-ci-dev feed for the .All metapackage in testing the runtime store.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -204,5 +204,17 @@
     </ItemGroup>
 
     <ZipArchive File="%(MetaPackageNupkg.FullPath)" SourceFiles="@(ArchiveFiles)" WorkingDirectory="@(MetaPackageNupkg->'$(TempDir)%(Filename)')" Overwrite="true" />
+
+    <PropertyGroup>
+      <TempPackageDir>$(TempDir)package\</TempPackageDir>
+      <PackagePublisherBinary>$(ToolsDir)\PackagePublisher\PackagePublisher.dll</PackagePublisherBinary>
+      <BuildToolsFeed>https://dotnet.myget.org/F/aspnetcore-tools/api/v2</BuildToolsFeed>
+      <PublishFeed>https://dotnet.myget.org/F/aspnetcore-ci-dev/</PublishFeed>
+    </PropertyGroup>
+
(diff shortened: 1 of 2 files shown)


build/repo.targets
```diff
@@ -204,5 +204,17 @@
...
+
+    <PropertyGroup>
+      <TempPackageDir>$(TempDir)package\</TempPackageDir>
+      <PackagePublisherBinary>$(ToolsDir)\PackagePublisher\PackagePublisher.dll</PackagePublisherBinary>
+      <BuildToolsFeed>https://dotnet.myget.org/F/aspnetcore-tools/api/v2</BuildToolsFeed>
+      <PublishFeed>https://dotnet.myget.org/F/aspnetcore-ci-dev/</PublishFeed>
+    </PropertyGroup>
+
+    <Copy SourceFiles="%(MetaPackageNupkg.FullPath)" DestinationFolder="$(TempPackageDir)" />
+    <Exec Command="$(ToolsDir)NuGet.exe install -ExcludeVersion -pre PackagePublisher -Source $(BuildToolsFeed) -out $(ToolsDir)"/>
```
kichalla: Why not use PackageReference? You can use this property to get the dll https://github.com/aspnet/BuildTools/blob/dev/src/PackagePublisher/build/PackagePublisher.props#L3

JunTaoLuo: Ah, forgot we had that, I was following what we had here: https://github.com/aspnet/Coherence-Signed/blob/dev/dnx.msbuild#L505-L517

JunTaoLuo:
Just realized this doesn't quite work. The reason being that coherence has already pushed the .All metapackage and here we are trying to push the same package to the same feed with the same version but different contents.

JunTaoLuo:
I see two alternatives:
A. we add an exclusion mechanism to Coherence that prevents pushing the .All package and push it later in the Coherence-PackageCache build. 
B. we change the order of the Coherence and Coherence-PackageCache builds and run Coherence after the runtime store builds are complete

The problem with A is that we'll be missing the .All package in the ci-dev feed until the Coherence-PackageCache completes. This means that the packages on the ci-dev feeds are not completely "coherent". The problem with B is that we'll add another 24 minutes to our overall build time. cc @ryanbrandenburg due to the impact on builds.

JunTaoLuo:
This will need a bit more though on how the builds flow. Closing for now.

PR #128, opened 2017-05-25
Fix command line args not working when using CreateDefaultBuilder

cnblogs-dudu:
This PR is to fix command line args not working when using `WebHost.CreateDefaultBuilder(args)`.

For example, when run `dotnet run --urls http://*:8003` with `WebHost.CreateDefaultBuilder(args)`, Kestrel server is still listening on port 5000.

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -169,11 +169,6 @@ namespace Microsoft.AspNetCore
                     }
 
                     config.AddEnvironmentVariables();
-
-                    if (args != null)
-                    {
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
This has been addressed in https://github.com/aspnet/MetaPackages/issues/221. Please feel free to reopen if there are additional functionalities required.

SuperJMN:
I think this is still not working. I'm passing `--urls=http://localhost:5001/' in the command line and it doesn't get it

Tratcher:
@SuperJMN this fix has not been released yet. You can try getting it from our myget.org feeds as shown on https://github.com/aspnet/home

PR #129, opened 2017-05-26
Target NETStandard2.0

JunTaoLuo:
Note a few caveats in this repo is a bit different. 

The only package we need to convert to NStandard2.0 is Microsoft.AspNetCore. M.A.All stays as netcoreapp2.0. 
The tests are a little misleading since there are functional and unit tests in the FunctionalTest project. I've added the net461 TFM to the FuntionalTests so that the unit tests are tested on both platforms. I'll file a follow up issue to split the functional and unit tests and also add coverage for .NET Framework variations for the functional tests.

### DIFF
--- a/src/Microsoft.AspNetCore/CertificateLoader.cs
+++ b/src/Microsoft.AspNetCore/CertificateLoader.cs
@@ -179,7 +179,13 @@ namespace Microsoft.AspNetCore
             {
                 var certificate = TryLoad(X509KeyStorageFlags.DefaultKeySet, out var error)
                     ?? TryLoad(X509KeyStorageFlags.UserKeySet, out error)
-                    ?? TryLoad(X509KeyStorageFlags.EphemeralKeySet, out error);
+#if NETCOREAPP2_0
+                    ?? TryLoad(X509KeyStorageFlags.EphemeralKeySet, out error)
+#elif NET461
+#else
+#error target frameworks need to be updated
+#endif
+                    ;
 
                 if (error != null)
                 {
--- a/test/Microsoft.AspNetCore.Tests/Microsoft.AspNetCore.Tests.csproj
+++ b/test/Microsoft.AspNetCore.Tests/Microsoft.AspNetCore.Tests.csproj
@@ -3,11 +3,12 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore/CertificateLoader.cs
```diff
@@ -178,8 +178,7 @@ public CertificateFileSource(ICertificateFileLoader certificateFileLoader)
             public override X509Certificate2 Load()
             {
                 var certificate = TryLoad(X509KeyStorageFlags.DefaultKeySet, out var error)
-                    ?? TryLoad(X509KeyStorageFlags.UserKeySet, out error)
-                    ?? TryLoad(X509KeyStorageFlags.EphemeralKeySet, out error);
```
JunTaoLuo: This isn't supported on in netstandard2.0 and is only available on netcoreapp2.0 cc @javiercn

test/Microsoft.AspNetCore.Tests/Microsoft.AspNetCore.Tests.csproj
```diff
@@ -3,11 +3,12 @@
...
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp2.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.All\Microsoft.AspNetCore.All.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore\Microsoft.AspNetCore.csproj" />
```
JunTaoLuo: Not sure why this referenced the .All metapackage. It is sufficient and makes more sense to use the Microsoft.AspNetCore mini metapackage instead.

HaoK: Yeah my bad, moved this back and forth between .All.Tests and just the core one, doesn't need a reference to all anymore

src/Microsoft.AspNetCore/CertificateLoader.cs
javiercn: You need to support ephemeral

JunTaoLuo: Will cross-compile

JunTaoLuo:
@Eilon @DamianEdwards We need to cross-compile Microsoft.AspNetCore mini-metapackage since we needed to load certificates with the EphemeralKeySet (which is only available on netcoreapp2.0) to ensure Azure scenarios work. @javiercn has more details.

PR #136, opened 2017-06-02
Faster runtime store build

JunTaoLuo:
Consolidating the individual package references into a separate project and referencing that in the runtime store project reduces the build time ~90%.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -32,6 +32,8 @@
     <TempDir>$(ArtifactsDir)temp\</TempDir>
     <ToolsDir>$(RepositoryRoot)tools\</ToolsDir>
     <DependencyBuildDirectory>$(RepositoryRoot).deps\build\</DependencyBuildDirectory>
+    <RuntimeStoreReferenceDirectory>$(ToolsDir)Build.RuntimeStore.References\</RuntimeStoreReferenceDirectory>
+    <RuntimeStoreReferencePackageDirectory>$(RuntimeStoreReferenceDirectory)bin\Release\</RuntimeStoreReferencePackageDirectory>
     <PrepareDependsOn>
       $(PrepareDependsOn);
       UpdateNuGetConfig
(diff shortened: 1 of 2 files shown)


natemcmaster:
I tried to run locally and this fails. Known issue?

> C:\Users\namc\.dotnet\x64\sdk\2.0.0-preview2-006215\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.CrossGen.targets(49,5): error : Unable to find resolved path for 'coreclr'. [C:\dev\aspnet\MetaPackages\src\Microsoft.AspNetCore.RuntimeStore\Microsoft.AspNetCore.RuntimeStore.csproj]

JunTaoLuo:
@natemcmaster yes, we pass in the platform via environment variable. set PACKAGE_CACHE_PLATFORM="x64/x86" and it should work

natemcmaster:
Do you know why it's so much faster? Seems too good to be true, or there's a really bad bug in `dotnet store` targets. I still can't get this to run locally to verify it still works

Also, can we add an error when PACKAGE_CACHE_PLATFORM is undefined?

build/repo.targets
```diff
@@ -64,10 +66,23 @@
       <OutputZipNoTimestamp>$(ArtifactsDir)Build.RS.$(OutputZipSufix).zip</OutputZipNoTimestamp>
     </PropertyGroup>
 
+    <!-- Build reference package -->
+    <Exec Command="dotnet restore" WorkingDirectory="$(RuntimeStoreReferenceDirectory)" />
+    <Exec Command="dotnet pack -c Release" WorkingDirectory="$(RuntimeStoreReferenceDirectory)" />
```
natemcmaster: FYI - you don't have to shell out to dotnet.exe to run this target.
```xml
<MSBuild Projects="$(RuntimeStoreReferenceDirectory)\Build.RuntimeStore.References.csproj"
    Targets="Pack"
    Properties="Configuration=Release" />
```

PR #127, opened 2017-05-25
Update manifest names

JunTaoLuo:
Addresses #124 to clarify which manifests are problematic when error message is displayed.

PR #143, opened 2017-06-06
Use file globbing to resolve manifests

JunTaoLuo:
Following up to https://github.com/aspnet/MetaPackages/pull/127.

Now that we have an updated CLI with the changes we needed, we can at least use file globbing to resolve the manifest names. This can go into preview2 but is also not necessary.

### DIFF
--- a/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
+++ b/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
@@ -11,8 +11,7 @@ Error if PublishWithAspNetCoreTargetManifest is set to true for standalone app
 -->
   <Target
     Name="PublishWithAspNetCoreTargetManifest"
-    BeforeTargets="Publish"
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
```diff
@@ -15,12 +11,19 @@ Error if PublishWithAspNetCoreTargetManifest is set to true for standalone app
 -->
   <Target
     Name="PublishWithAspNetCoreTargetManifest"
-    BeforeTargets="Publish"
-    DependsOnTargets="PrepareForPublish"
+    AfterTargets="PrepareForPublish"
```
JunTaoLuo: :up::date: we need to run this target earlier than we have been to ensure TargetManifestFiles is set before all dotnet/sdk targets that rely on it.

JunTaoLuo:
Closed in https://github.com/aspnet/MetaPackages/commit/35ec06975d3b3738e83c6381d587c7c1d9e2eaf1

PR #150, opened 2017-06-08
Fix PackageCache builds

JunTaoLuo:
No symbols are generated for macOS

Add environment variable to pass additional arguments to dotnet store


JunTaoLuo:
Merging now to fix build

PR #151, opened 2017-06-08
Re-enable symbols generation on x86

JunTaoLuo:
We temporarily disabled the generation of symbols for runtime stores on x86 due to missing DiaSymReader binaires. The proposal here is to download the x86 shared runtime and add the required binary to the PATH to ensure symbol generation on x86 succeeds.

We need this for preview2


tools/InstallSharedFrameworkx86.ps1
```diff
@@ -0,0 +1,28 @@
...
+    [string]$installationDirectory
+)
+
+$sharedRuntimeChannel = "master"
+if ($env:KOREBUILD_DOTNET_SHARED_RUNTIME_CHANNEL)
+{
+    $sharedRuntimeChannel = $env:KOREBUILD_DOTNET_SHARED_RUNTIME_CHANNEL
+}
+
+function InstallSharedRuntime([string] $version, [string] $channel, [string] $installDir)
```
JunTaoLuo: Copied most of this script from KoreBuild.ps1

muratg:
This looks good to me overall, but I'm a bit surprised we don't already have the corresponding shared runtime on the agent (so having to download it.) Are we making sure it's the correct version of the shared runtime?

I'll leave the sign-off to other folks in the reviewers list.

JunTaoLuo:
@muratg Our builds always [installs](https://github.com/aspnet/KoreBuild/blob/dev/build/KoreBuild.ps1#L26) the x64 version of the shared runtime and cli. Symbol generation on x86 seems to be a special case so I'm re-running part of KoreBuild.ps1 that installs shared runtimes. I resolve the version from the current CLI and use the same one to install the x86 runtime so it should be kept in sync with the rest of the build.

muratg:
@JunTaoLuo With the recent changes in the CI, the shared runtime version coming from the CLI may not be matching the shared runtime we picked up through our Integration ("CoreCLR") build. Maybe another mechanism is required?

cc @mikeharder

JunTaoLuo:
@muratg I spoke with @mikeharder and this mechanism is fine. Here, I'm using dotnet --info to figure out the version of the current shared framework host which is auto-rolled forward to the latest version installed. When we install a newer version of the shared framework, I have verified we get the latest version instead of the one that was bundled with the CLI.

PR #152, opened 2017-06-11
Even shorter notation to build and run a Webhosts

faniereynders:
Added support for allowing a shorthand notation to build and run a Webhost with the default configuration including a typed Startup class

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -191,5 +191,40 @@ namespace Microsoft.AspNetCore
 
             return builder;
         }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WebHostBuilder"/> class with pre-configured defaults using typed Startup
+        /// </summary>
+        /// <typeparam name="T">Specify the startup type to be used by the web host.</typeparam>
+        /// <param name="args">The command line args.</param>
+        /// <remarks>
+        ///   The following defaults are applied to the returned <see cref="WebHostBuilder"/>:
+        ///     use Kestrel as the web server,
+        ///     set the <see cref="IHostingEnvironment.ContentRootPath"/> to the result of <see cref="Directory.GetCurrentDirectory()"/>,
+        ///     load <see cref="IConfiguration"/> from 'appsettings.json' and 'appsettings.[<see cref="IHostingEnvironment.EnvironmentName"/>].json',
+        ///     load <see cref="IConfiguration"/> from User Secrets when <see cref="IHostingEnvironment.EnvironmentName"/> is 'Development' using the entry assembly,
+        ///     load <see cref="IConfiguration"/> from environment variables,
+        ///     configures the <see cref="ILoggerFactory"/> to log to the console and debug output,
+        ///     enables IIS integration,
+        ///     enables the ability for frameworks to bind their options to their default configuration sections,
+        ///     adds the developer exception page when <see cref="IHostingEnvironment.EnvironmentName"/> is 'Development'
+        ///     and sets the startup class as the typed defined in T.
+        /// </remarks>
+        /// <returns>The initialized <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder CreateDefaultBuilder<T>(string[] args) where T : class
+        {
+            return WebHost.CreateDefaultBuilder(args)
+                .UseStartup<T>();
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


faniereynders:
@davidfowl?

muratg:
@JunTaoLuo @davidfowl could you guys take a look?

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -191,5 +191,40 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
 
             return builder;
         }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WebHostBuilder"/> class with pre-configured defaults using typed Startup
+        /// </summary>
+        /// <typeparam name="T">Specify the startup type to be used by the web host.</typeparam>
```
JunTaoLuo: I would prefer using the same language as `UseStartup` https://github.com/aspnet/Hosting/blob/70a3017b12b4e9648f65acea2c9ad37e383763b0/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs#L77

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -191,5 +191,40 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
...
+        /// <param name="args">The command line args.</param>
```
JunTaoLuo: `summary` first, then `remarks`, then `params` and lastly `returns`.

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -191,5 +191,40 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
...
+        /// <remarks>
+        ///   The following defaults are applied to the returned <see cref="WebHostBuilder"/>:
+        ///     use Kestrel as the web server,
+        ///     set the <see cref="IHostingEnvironment.ContentRootPath"/> to the result of <see cref="Directory.GetCurrentDirectory()"/>,
+        ///     load <see cref="IConfiguration"/> from 'appsettings.json' and 'appsettings.[<see cref="IHostingEnvironment.EnvironmentName"/>].json',
+        ///     load <see cref="IConfiguration"/> from User Secrets when <see cref="IHostingEnvironment.EnvironmentName"/> is 'Development' using the entry assembly,
+        ///     load <see cref="IConfiguration"/> from environment variables,
```
JunTaoLuo: We also load configuration from supplied command line args.

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -191,5 +191,40 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
...
+            return WebHost.CreateDefaultBuilder(args)
+                .UseStartup<T>();
+        }
+
+        /// <summary>
+        /// Builds an Microsoft.AspNetCore.Hosting.IWebHost which hosts a web application and
+        /// runs a web application and block the calling thread until host shutdown.
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebHostBuilder"/></param>
+        public static void BuildAndRun(this IWebHostBuilder builder)
```
JunTaoLuo: I'm not sure if we want to add this extension. It doesn't save that much compared to the original `Build().Run()`. Though if we were to add this extension, we would also need to add `BuildAndRunAsync()`.

test/TestSites/CreateDefaultBuilderApp/Startup.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Microsoft.AspNetCore.Builder;
```
JunTaoLuo: Add copyright header

test/TestSites/CreateDefaultBuilderApp/Program.cs
```diff
@@ -36,6 +36,11 @@ static void Main(string[] args)
                     });
                 })
                 .Build().Run();
+
+            Console.ReadKey();
```
JunTaoLuo: This doesn't add any coverage for the new overload other than for compilation. Create a copy of this app that uses `CreateDefaultBuilder<Startup>` and add an test like https://github.com/aspnet/MetaPackages/blob/dev/test/Microsoft.AspNetCore.FunctionalTests/WebHostFunctionalTests.cs#L51 instead.

JunTaoLuo:
Thanks @faniereynders for the contribution! Your changes were merged in https://github.com/aspnet/MetaPackages/commit/07bf8d02a054e37461c8ccf50dc95ee594556c62.

PR #164, opened 2017-06-19
Remove SQLitePCLRaw workaround

natemcmaster:
No longer required with SQLitePCLRaw 1.1.6.

See https://github.com/aspnet/Microsoft.Data.Sqlite/pull/375 and https://github.com/aspnet/EntityFramework/pull/8890


natemcmaster:
Blocked on https://github.com/aspnet/EntityFramework/pull/8890 - this update breaks some query behavior in EFCore.SQLite.

natemcmaster:
Closing. This version of SQLite bundled in this update appears to includes a regression in LEFT JOIN. Let's not update until we have a fix or a different bundled version.

Issue #161, opened 2017-06-16
Publish not trimming for -final packages


Tratcher:
The -final .All package includes both versioned and final manifest files. E.g. aspnetcore-store-2.0.0-preview2-25756-linux-x64.xml and aspnetcore-store-2.0.0-preview2-final-linux-x64.xml.
The PublishWithAspNetCoreTargetManifest.targets file references the versioned manifests. These all reference versioned packages, which do not match the -final packages referenced by your app, so nothing gets trimmed.

@muratg @Eilon Preview2 blocking

Eilon:
Moving to RTM milestone because we did the workaround for Preview2.

JunTaoLuo:
This issue is masked in 2.0.0 by https://github.com/aspnet/MetaPackages/pull/143 which means even without any changes, trimming should work correctly for the -final package. However, we still need to fix this to remove the redundant manifests in the package. Also, this will break new builds of preview2. Note that the issue actually lies with the build scripts in Coherence-Signed, there won't be changes in this directory.

JunTaoLuo:
The trimming issue has been fixed as part of https://github.com/aspnet/MetaPackages/pull/167 in preview2. The duplication is fixed by https://github.com/aspnet/Coherence-Signed/pull/588 in dev.

PR #167, opened 2017-06-19
Remove ViewCompilation from publish manifest

JunTaoLuo:
We are running into an issue where the store generated on windows and non-windows are different. The windows stores (x86 and x64) do not contain the M.A.Mvc.Razor.ViewCompilation while the non windows store (macOS and linux) do not. Our assumption that the stores on all platforms are the same is now broken so our approach of trimming by all manifests no longer works. We need to react accordingly in the M.A.All metapackage target and the proposal is to remove the ViewCompilation entry from the manifest so it will never be trimmed.

This would resolve https://github.com/aspnet/MvcPrecompilation/issues/146, note that after preview2 we plan to remove ViewCompilation from the runtime store so this will not be an issue for the next release.

I also ported the fix from https://github.com/aspnet/MetaPackages/pull/143, this helps us  resolve https://github.com/aspnet/MetaPackages/issues/161 without having to manually update the target file in the M.A.All package.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -166,6 +166,9 @@
     <!--Drop a nuspec file in artifacts for packing zip files into a nupkg-->
     <Copy SourceFiles="$(RepositoryRoot)build\Build.RS.nuspec" DestinationFolder="$(ArtifactsDir)" Condition="'$(OSPlatform)'=='Linux'" />
     <WriteLinesToFile File="$(ArtifactsDir)version.txt" Lines="$(VersionPrefix)-$(VersionSuffix)" Overwrite="true" Condition="'$(OSPlatform)'=='Linux'" />
+
+    <!-- Preview 2 Workaround: remove Microsoft.AspNetCore.Mvc.Razor.ViewCompilation from publish manifest -->
+    <Exec Command="sed -i -e '/microsoft.aspnetcore.mvc.razor.viewcompilation/d' $(ArtifactsDir)%(PackageStoreManifestFiles.TimestampDestinationFile)" Condition="'$(OS)' != 'Windows_NT'" />
   </Target>
 
   <Target Name="ConvertZipToTGZ">
--- a/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
+++ b/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
@@ -3,10 +3,6 @@
     <PublishWithAspNetCoreTargetManifest Condition="'$(PublishWithAspNetCoreTargetManifest)'=='' and '$(RuntimeIdentifier)'=='' and '$(RuntimeIdentifiers)'=='' and '$(PublishableProject)'=='true'">true</PublishWithAspNetCoreTargetManifest>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(PublishWithAspNetCoreTargetManifest)'=='true'">
-    <TargetManifestFiles>$(TargetManifestFiles);$(MSBuildThisFileDirectory)aspnetcore-store-__MANIFEST_VERSION__-win7-x64.xml;$(MSBuildThisFileDirectory)aspnetcore-store-__MANIFEST_VERSION__-win7-x86.xml;$(MSBuildThisFileDirectory)aspnetcore-store-__MANIFEST_VERSION__-osx-x64.xml;$(MSBuildThisFileDirectory)aspnetcore-store-__MANIFEST_VERSION__-linux-x64.xml</TargetManifestFiles>
-  </PropertyGroup>
-
 <!--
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
```diff
@@ -15,12 +11,21 @@ Error if PublishWithAspNetCoreTargetManifest is set to true for standalone app
...
       Condition="'$(RuntimeIdentifier)'!='' or '$(RuntimeIdentifiers)'!=''" />
+
+    <ItemGroup>
+      <WindowsAspNetCoreTargetManifestFiles Include="$(MSBuildThisFileDirectory)aspnetcore-store-*win*.xml"/>
+      <NonWindowsAspNetCoreTargetManifestFiles Include="$(MSBuildThisFileDirectory)aspnetcore-store-*.xml" Exclude="$(MSBuildThisFileDirectory)aspnetcore-store-*win*.xml"/>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <TargetManifestFiles Condition="'$(OS)' == 'Windows_NT'">$(TargetManifestFiles);@(WindowsAspNetCoreTargetManifestFiles)</TargetManifestFiles>
+      <TargetManifestFiles Condition="'$(OS)' != 'Windows_NT'">$(TargetManifestFiles);@(NonWindowsAspNetCoreTargetManifestFiles)</TargetManifestFiles>
```
natemcmaster: Wouldn't this mean we can no longer publish Linux apps from Windows and vice-versa?

build/repo.targets
```diff
@@ -208,6 +211,11 @@
     <SetEnvironmentVariable Variable="PATH" Value="$(PATH);$(DiaSymReaderDirectory)" />
   </Target>
 
+  <Target Name="_RemoveViewCompilationFromManifest">
+    <Exec Command="powershell.exe -command &quot;Get-Content $(ManifestFileWithViewCompilation) | Where-Object {$_ -inotmatch ''} | Set-Content $(ManifestFileWithViewCompilation)&quot;" Condition="'$(OS)' == 'Windows_NT'"/>
```
JunTaoLuo: Only need to remove on macOS and linux.

mikeharder:
Are the changes in `Microsoft.AspNetCore.All.csproj` and `PublishWithAspNetCoreTargetManifest.targets` required for Preview2?  They seem riskier than the core change in `repo.targets`.

natemcmaster:
Why don't we just remove this? 

https://github.com/aspnet/MetaPackages/blob/rel/2.0.0-preview2/build/common.props#L133

build/repo.targets
```diff
@@ -166,6 +166,9 @@
     <!--Drop a nuspec file in artifacts for packing zip files into a nupkg-->
     <Copy SourceFiles="$(RepositoryRoot)build\Build.RS.nuspec" DestinationFolder="$(ArtifactsDir)" Condition="'$(OSPlatform)'=='Linux'" />
     <WriteLinesToFile File="$(ArtifactsDir)version.txt" Lines="$(VersionPrefix)-$(VersionSuffix)" Overwrite="true" Condition="'$(OSPlatform)'=='Linux'" />
+
+    <!-- Preview 2 Workaround: remove Microsoft.AspNetCore.Mvc.Razor.ViewCompilation from publish manifest -->
+    <Exec Command="sed -i -e '/microsoft.aspnetcore.mvc.razor.viewcompilation/d' $(ArtifactsDir)%(PackageStoreManifestFiles.TimestampDestinationFile)" Condition="'$(OS)' != 'Windows_NT'" />
```
pranavkm: Can't we just remove the package reference? Why mess with the build output?

JunTaoLuo: I thought this may have been pulled in transitively.

pranavkm: Nope, it's a standalone top-level package. Should be safe to just remove the package reference.

JunTaoLuo: Actually wouldn't we need to make it a FullMetaPackagePackageNoRuntimeStoreReference? we want it to be a transitive dependency I assume? But not in the runtime store.

JunTaoLuo: But we'd need to remove the PrivateAssets=All metadata

pranavkm: Spoke to @JunTaoLuo offline. Seems like this is the only way right now to get a reference in the package, but not have the package in the store.

JunTaoLuo: To clarify, we are not removing the package from the store, we just never use the one in the store by forcing ViewCompilation to be bin deployed.

JunTaoLuo: I feel the alternative which is to remove it from the store completely is doable, but I'd want to test it more since it's slightly more riskier. Figured better to unblock the build with this now.

Issue #155, opened 2017-06-14
Use artifacts from Coherence-Signed for installer generation and fallback to myget if it's not available


JunTaoLuo:
Follow up to https://github.com/aspnet/MetaPackages/pull/154#issuecomment-308511071. Currently this depends on Coherence-Signed, we want to be able to run this locally with unsigned bits when building from source.

JunTaoLuo:
The debian installer generation is currently standalone and does not depend on Coherence-Signed as it downloads the runtime store archives from myget instead of the artifacts from Coherence-Signed. 

This however leads to a problem in our builds where the latest archives are not being used to build the installers due to the indexing delays of myget. For example in the first 25794 build of the installers, the 25778 archives were used: http://aspnetci/viewLog.html?buildId=244765&buildTypeId=Coherence_RuntimeStoreInstallers&tab=buildLog&state=&expand=all#_focus=327. 

Instead, the work that needs to be done here is to add a mechanism to retrieve the archives from Coherence-Signed but fallback to myget if it's not available so the installer builds remain standalone.

PR #174, opened 2017-06-23
Use specified signed runtime store archives when available for installers

JunTaoLuo:
Addresses https://github.com/aspnet/MetaPackages/issues/155. 

This will allow our builds to pick up the latest archives from Coherence-Signed instead of relying on myget which tends to pick up older builds due to lag in indexing.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -322,11 +322,17 @@
     <Error Text="Docker host must be using Linux containers." Condition="'$(DockerHostOS)' != 'linux'"/>
 
     <!-- Download runtime store archives -->
-    <Exec Command="dotnet restore --packages $(RuntimeStoreArchiveDirectory)" WorkingDirectory="$(RestoreRuntimeStoreDirectory)" />
+    <Message Text="Local runtime store signed artifacts not specified via RUNTIMESTORE_SIGNED_ARTIFACTS_PATH, falling back to artifacts availabe on myget." Importance="high" Condition="'$(RUNTIMESTORE_SIGNED_ARTIFACTS_PATH)' == ''"/>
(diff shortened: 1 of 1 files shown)


build/repo.targets
```diff
@@ -322,11 +322,17 @@
     <Error Text="Docker host must be using Linux containers." Condition="'$(DockerHostOS)' != 'linux'"/>
 
     <!-- Download runtime store archives -->
-    <Exec Command="dotnet restore --packages $(RuntimeStoreArchiveDirectory)" WorkingDirectory="$(RestoreRuntimeStoreDirectory)" />
+    <Message Text="Local runtime store signed artifacts not specified via RUNTIMESTORE_SIGNED_ARTIFACTS_PATH, falling back to artifacts availabe on myget." Importance="high" Condition="'$(RUNTIMESTORE_SIGNED_ARTIFACTS_PATH)' == ''"/>
+    <Exec Command="dotnet restore --packages $(RuntimeStoreArchiveDirectory)" WorkingDirectory="$(RestoreRuntimeStoreDirectory)" Condition="'$(RUNTIMESTORE_SIGNED_ARTIFACTS_PATH)' == ''"/>
```
natemcmaster: FYI, dotnet-restore can be written as an MSBuild command instead of using dotnet.exe.
```xml
<MSBuild 
    Projects="$(ToolsDir)tools\RestoreRuntimeStore\RestoreRuntimeStore.csproj"
    Properties="RestorePackagesPath=$(RuntimeStoreArchiveDirectory)" />
```

natemcmaster:
Not sure exactly how this will work. Isn't Build.RS.nupkg only on the signed feeds (aspnetcore-dev), and not aspnetcore-ci-dev?

JunTaoLuo:
The Linux agent will have the drop share mounted and we will be grabbing the artifacts from Coherence-Signed directly instead of going through myget. This way we can avoid mismatches caused by myget delays.

natemcmaster:
Would it make more sense to use an Artifact Dependency between the builds to avoid the dotnet-restore altogether?

JunTaoLuo:
The dotnet-restore is only done when the drop-share isn't available. Artifact dependencies will work too, but we need to update the mechanism to fallback to myget when it's not available. Unless we don't want to preserve self contained builds without access to drop share.

Issue #175, opened 2017-06-23
ASP.NET TargetManifestFiles should use $(SelfContained) property


eerhardt:
See
https://github.com/aspnet/MetaPackages/blob/35ec06975d3b3738e83c6381d587c7c1d9e2eaf1/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets#L3
and
https://github.com/aspnet/MetaPackages/blob/35ec06975d3b3738e83c6381d587c7c1d9e2eaf1/src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets#L19

These conditions shouldn't be using `'$(RuntimeIdentifier)'=='' and '$(RuntimeIdentifiers)'==''`, but instead it should use the new `$(SelfContained)` property instead.

In .NET Core 2.0, we now have a 3rd app model: Rid-specific, Shared Framework apps (sometimes referred to as a "low-fat app").  This means, I can publish an app that runs on the Shared Framework, but it only uses the `win-x64` (or any other RID) assets.  It isn't truly "portable" meaning it will run on any platform.  This trims down on the size of the app, as it doesn't need to publish assets for other platforms that it won't run on.  See https://github.com/dotnet/sdk/pull/1053.

Thus, if someone tried publishing a Rid-specific, Shared Framework ASP.NET app and wanted to use the ASP.NET runtime store, these checks wouldn't be correct.

To fix this, simply change the RuntimeIdentifier checks to use SelfContained instead.  Target manifest files shouldn't be used with SelfContained apps.  But they should default to be used when the app isn't SelfContained.

/cc @JunTaoLuo @nguerrera

PR #177, opened 2017-06-24
Use new SelfContained property

JunTaoLuo:
Addresses https://github.com/aspnet/MetaPackages/issues/175. 

We will inspect this property instead of RuntimeIdentifier/RuntimeIdentifiers to determine whether the TargetManifestFiles should be updated. This new property is set in the sdk: https://github.com/dotnet/sdk/blob/43090565482d9a10481ff2298fb168fc819fa430/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.RuntimeIdentifierInference.targets#L107-L110

Still needs to be tested before being reviewed.


JunTaoLuo:
Tested this with SelfContained app (RID specific, aka standalone), RID specific portable and portable apps. Ready to merge.

PR #178, opened 2017-06-27
Workaround issue caused by importing Internal.AspNetCore.Sdk into repo.targets

natemcmaster:
Internal.AspnetCore.Sdk should only be used in \*.csproj files, not repo.targets. Because repo.targets imports common.props, it ends up with a reference with this package, which is causing build failures due to changes in Internal.AspNetCore.BuildTools.Tasks.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -1,6 +1,11 @@
 <Project>
   <Import Project="common.props" />
 
+  <ItemGroup>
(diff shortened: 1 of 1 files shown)


natemcmaster:
It would be better to refactor common.props and to figure out what repo.targets actually needs from common.props.

Issue #156, opened 2017-06-15
Automate package name generation in debian_config.json


JunTaoLuo:
Currently this is hardcoded in https://github.com/aspnet/MetaPackages/blob/efbc95386815b8a77f912eef3e36a7509ec7a1d3/tools/debian_config.json#L8. We should automate this so we don't have to manually update this every release.

JunTaoLuo:
Work will actually be done in Universe.

JunTaoLuo:
Work completed in rel/2.0.3

PR #186, opened 2017-06-29
Installer generation updates

JunTaoLuo:
Use the new `JsonPeek` task to parse common installer parameters from debian_config.json to reduce duplication. Tested manually.

### DIFF
--- a/tools/debian_config.json
+++ b/tools/debian_config.json
@@ -5,7 +5,7 @@
     "package_name": "aspnetcore-store",
     "install_root": "/usr/share/dotnet",
 
-    "short_description": "Microsoft ASP.NET Core Runtime Package Store 2.0.0 - Preview 3",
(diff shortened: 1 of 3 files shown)


tools/debian_config.json
```diff
@@ -5,7 +5,7 @@
     "package_name": "aspnetcore-store",
     "install_root": "/usr/share/dotnet",
 
-    "short_description": "Microsoft ASP.NET Core Runtime Package Store 2.0.0 - Preview 3",
+    "short_description": "Microsoft ASP.NET Core Runtime Package Store 2.0.0",
```
natemcmaster: It would be good to find a way to pull the version number from version.props. Seems very possible that we'll forget to update this in the future.

JunTaoLuo: Tracked in https://github.com/aspnet/MetaPackages/issues/156

Issue #172, opened 2017-06-21
win-arm64\native\sni.dll not trimmed during publish


mikeharder:
# Repro Steps
1. `dotnet new web`
2. `dotnet publish`

# Expected
publish dir contains only app binaries/config

# Actual
```
runtimes\win-arm64\native\sni.dll
```

# Dependency Versions
* CLI: https://dotnetcli.azureedge.net/dotnet/Sdk/2.0.0-preview2-006494/dotnet-sdk-2.0.0-preview2-006494-win-x64.zip

PR #188, opened 2017-07-03
Create common manifest file

JunTaoLuo:
Addresses: https://github.com/aspnet/MetaPackages/issues/172

This will trim additional packages such as runtime.win-arm64.runtime.native.system.data.sqlclient.sni

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -175,8 +175,24 @@
     <Copy SourceFiles="$(RepositoryRoot)build\Build.RS.nuspec" DestinationFolder="$(ArtifactsDir)" Condition="'$(OSPlatform)'=='Linux'" />
     <WriteLinesToFile File="$(ArtifactsDir)version.txt" Lines="$(VersionPrefix)-$(VersionSuffix)" Overwrite="true" Condition="'$(OSPlatform)'=='Linux'" />
 
-    <!-- Preview 2 Workaround: remove Microsoft.AspNetCore.Mvc.Razor.ViewCompilation from publish manifest -->
-    <Exec Command="sed -i -e '/microsoft.aspnetcore.mvc.razor.viewcompilation/d' $(ArtifactsDir)%(PackageStoreManifestFiles.TimestampDestinationFile)" Condition="'$(OS)' != 'Windows_NT'" />
+    <!-- Add a common manifest for package trimming -->
+    <MSBuild Projects="$(ProjectPath)" Targets="CreateCommonManifest" Condition="'$(OSPlatform)' == 'Windows' AND '$(PACKAGE_CACHE_PLATFORM)' == 'x64'"/>
+  </Target>
+
+  <Target Name="CreateCommonManifest">
+    <PropertyGroup>
+      <CommonManifestFileName>aspnetcore-store-$(TimestampVersion)-common.xml</CommonManifestFileName>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageToTrim Include="runtime.win-arm64.runtime.native.system.data.sqlclient.sni" />
+    </ItemGroup>
+
+    <MSBuild Projects="$(RuntimeStoreReferenceDirectory)\Build.RuntimeStore.References.csproj" Targets="GetPackageDefinitions" >
+      <Output TaskParameter="TargetOutputs" ItemName="_PackageDefinitions" />
+    </MSBuild>
+
+    <RepoTasks.CreateCommonManifest DestinationFilePath="$(ArtifactsDir)$(CommonManifestFileName)" PackageDefinitions="@(_PackageDefinitions)" Packages="@(PackageToTrim)"/>
   </Target>
 
   <Target Name="ConvertZipToTGZ">
--- a/build/tasks/RepoTasks.csproj
+++ b/build/tasks/RepoTasks.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+   <Import Project="$(RepoTasksSdkPath)\Sdk.props" Condition="'$(RepoTasksSdkPath)' != '' "/>
+
+   <PropertyGroup>
+       <TargetFramework>netcoreapp2.0</TargetFramework>
+   </PropertyGroup>
+
+  <Import Project="$(RepoTasksSdkPath)\Sdk.targets" Condition="'$(RepoTasksSdkPath)' != '' "/>
+</Project>
--- a/build/tasks/CreateCommonManifest.cs
+++ b/build/tasks/CreateCommonManifest.cs
@@ -1,0 +1,67 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Xml;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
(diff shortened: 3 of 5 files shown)


build/repo.targets
```diff
@@ -175,8 +175,23 @@
     <Copy SourceFiles="$(RepositoryRoot)build\Build.RS.nuspec" DestinationFolder="$(ArtifactsDir)" Condition="'$(OSPlatform)'=='Linux'" />
     <WriteLinesToFile File="$(ArtifactsDir)version.txt" Lines="$(VersionPrefix)-$(VersionSuffix)" Overwrite="true" Condition="'$(OSPlatform)'=='Linux'" />
 
-    <!-- Preview 2 Workaround: remove Microsoft.AspNetCore.Mvc.Razor.ViewCompilation from publish manifest -->
-    <Exec Command="sed -i -e '/microsoft.aspnetcore.mvc.razor.viewcompilation/d' $(ArtifactsDir)%(PackageStoreManifestFiles.TimestampDestinationFile)" Condition="'$(OS)' != 'Windows_NT'" />
+    <!-- Add a common manifest for package trimming -->
```
JunTaoLuo: The previous workaround is no longer needed after https://github.com/aspnet/MvcPrecompilation/issues/143. We only need to generate the common manifest once on the windows machines so I'm arbitrarily picking the win7-x64 builds.

build/repo.targets
```diff
@@ -175,8 +175,23 @@
...
+  </Target>
+
+  <Target Name="CreateCommonManifest">
+    <PropertyGroup>
+      <CommonManifestFileName>aspnetcore-store-$(TimestampVersion)-common.xml</CommonManifestFileName>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageToTrim Include="runtime.win-arm64.runtime.native.system.data.sqlclient.sni" />
+      <ProjectAssetsJson Include="$(RuntimeStoreReferenceDirectory)**\project.assets.json" />
```
JunTaoLuo: This step uses the restored lock file from tools/Build.RuntimeStore.References to determine the versions of the the packages that goes into the store. This project is guaranteed to be restored during the runtime store generation step.

build/tasks/CreateCommonManifestTask.cs
```diff
@@ -0,0 +1,72 @@
...
+            var packages = projectAssets
+                    .SelectToken("$.targets.*") // target/.NETCoreApp,Version=v2.0
+                    .Children().Select(s => ((JProperty)s).Name); // Package.Name/Version
+
+            var xmlDoc = new XmlDocument();
+            var packagesElement = xmlDoc.CreateElement("StoreArtifacts");
+
+            foreach (var package in Packages.Select(i => i.ItemSpec))
+            {
+                var packageVersion = packages
```
JunTaoLuo: There may be cleaner ways of doing this. along with parsing out the Package.Name/Version from the project.assets.json lines 40-42

build/tasks/CreateCommonManifestTask.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.IO;
+using System.Linq;
+using System.Xml;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Newtonsoft.Json.Linq;
+
+namespace RepoTasks
+{
+    public class CreateCommonManifestTask : Task
```
natemcmaster: Can you add a description of what this task does (for posterity and the next guy who has to edit this)?

build/repo.targets
```diff
@@ -175,8 +175,23 @@
...
+    </ItemGroup>
+
+    <Error Text="@(ProjectAssetsJson->Count()) project.assets.json file(s) found." Condition="'@(ProjectAssetsJson->Count())' != 1" />
+
+    <CreateCommonManifestTask DestinationFilePath="$(ArtifactsDir)$(CommonManifestFileName)" ProjectAssetsJsonFilePath="@(ProjectAssetsJson)" Packages="@(PackageToTrim)"/>
```
natemcmaster: Nit: use the task namespace when using repo tasks. `<RepoTasks.CreateCommonManifest ...`  Makes it more obvious where this task comes from.

build/tasks/CreateCommonManifestTask.cs
natemcmaster: Also, MSBuild convention is to leave off the `Task` name.  `public class CreateCommonManifest : Task`

build/dependencies.props
```diff
@@ -5,7 +5,9 @@
     <CoreFxVersion>4.4.0-*</CoreFxVersion>
     <DotnetDebToolVersion>2.0.0-preview2-*</DotnetDebToolVersion>
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
+    <JsonNetInMSBuildVersion>9.0.1</JsonNetInMSBuildVersion>
     <MoqVersion>4.7.49</MoqVersion>
+    <MsBuildPackageVersions>15.1.1012</MsBuildPackageVersions>
```
natemcmaster: Latest MSBuild is 15.3.0.

build/tasks/RepoTasks.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MsBuildPackageVersions)" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="$(MsBuildPackageVersions)" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MsBuildPackageVersions)" PrivateAssets="All" />
+    <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetInMSBuildVersion)" />
+  </ItemGroup>
+</Project>
```
natemcmaster: Nit: add trailing newlines.

build/dependencies.props
JunTaoLuo: The current versions we use in our repos is 15.1.1012, see https://github.com/search?utf8=%E2%9C%93&q=org%3Aaspnet+filename%3Adependencies.props+MsBuildPackageVersions&type=Code. Is there a reason why we can use a higher version here or does this need to be updated everywhere?

natemcmaster: Doesn't really matter as these are reference assemblies. MSBuild resolves this assembly to whatever version it carried. But since you've pointed it out, we should upgrade the other tools in our org.

natemcmaster: Btw, seeing this makes me think it would be better if KoreBuild set common properties and package references so that each RepoTasks.csproj didn't need it....

JunTaoLuo: That would be nice.

natemcmaster: https://github.com/aspnet/KoreBuild/pull/268

JunTaoLuo:
:up::date:

build/repo.targets
```diff
@@ -1,6 +1,8 @@
-<Project>
+<Project Sdk="Microsoft.NET.Sdk">
```
natemcmaster: This is going to cause MSBuild warnings about double imports of Microsoft.Common.targets. We shouldn't be using Microsoft.NET.Sdk from repo.targets.

PR #206, opened 2017-07-26
Rename store zip file names

JunTaoLuo:
Following up with offline discussions. We should probably rename the runtime store zip files to match the installer names. This is not critical since the artifacts are only used internally and not intended to be used by customers. Also in this PR is the removal of nuspec and version.txt files used to build the Build.RS nukpg which we no longer need. Once we are happy with this change, I'll need to update Coherence-Signed and SetupCrossGen to look for the new file names.


JunTaoLuo:
@anurse We also decided to change the naming scheme for the lzma archives so it conforms to the other artifacts picked up by the CLI. This should be safe right?

anurse:
This is for post-2.0? It would likely affect how the CLI pulls the LZMA, they'd need to use the updated name. /cc @livarcocc

JunTaoLuo:
Yes this is post 2.0. They are aware of this change and will be reacting to this change. (In fact, I believe they already have the reaction PR)

PR #208, opened 2017-07-27
ProjectPath is no longer set, use MSBuildThisFileFullPath instead

JunTaoLuo:
This should fix coherence-signed.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -161,7 +161,7 @@
       <NoTimestampDepsFiles Include="$(RuntimeStoreZipNoTimestampDir)additionalDeps\**\*"/>
     </ItemGroup>
 
-    <MSBuild Projects="$(ProjectPath)" Targets="_RemoveTimestampFromDepsFile" Properties="DepsFile=%(NoTimestampDepsFiles.FullPath)" />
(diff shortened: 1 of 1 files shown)


build/repo.targets
```diff
@@ -246,14 +246,8 @@
   </Target>
 
   <Target Name="_RemoveTimestampFromDepsFile">
-    <GetOSPlatform>
```
JunTaoLuo: Seems like this task was also missing after the bootstrapper update but I noticed it could be simplified away so I removed it.

natemcmaster: It's missing because `dotnet msbuild KoreBuild.proj` and `dotnet msbuild repo.targets` are not the same. Building repo.targets directly doesn't import the build tasks assembly.

FYI - MSBuild 15.3 has a new [helper method](https://github.com/Microsoft/msbuild/blob/v15.3.407.29267/src/Build/Evaluation/IntrinsicFunctions.cs#L431-L439).
`Condition=" $([MSBuild]::IsOSPlatform('OSX')) OR $([MSBuild]::IsOSPlatform('Windows')) OR $([MSBuild]::IsOSPlatform('Linux'))"`

JunTaoLuo: Ah, well it's not really needed at all so it should be removed regardless.

JunTaoLuo:
:up::date:

PR #210, opened 2017-08-01
Wire up the IncludeScopes configuration option, closes #204

### LINKED ISSUE #204: IncludeScopes in config doesn't work in the template
See:
https://github.com/aspnet/Logging/issues/643

Please fix before RTM

NinoFloris:
This option should affect the ConsoleLogger and isn't today

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -179,8 +179,11 @@ namespace Microsoft.AspNetCore
                 })
                 .ConfigureLogging((hostingContext, logging) =>
                 {
-                    logging.AddConfiguration(hostingContext.Configuration.GetSection("Logging"));
-                    logging.AddConsole();
+                    var loggingConfig = hostingContext.Configuration.GetSection("Logging");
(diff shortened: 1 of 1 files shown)


Eilon:
@muratg - who should review this?

muratg:
Assigning to @JunTaoLuo for review.

JunTaoLuo:
This looks odd. We pass in the configuration to the `ILoggerBuilder` but it does not flow the `IncludeScopes` configuration to the `ConsoleLogger`. I feel that the configuration of our loggers needs an update instead of this change. cc @BrennanConroy @pakrym thoughts?

pakrym:
This is caused by bigger issue and this is not the right solution.

Issue is that we need a general way to wire logger providers `ConfigureOptions` to `IConfiguration` instances assigned to logging configuration.

pakrym:
Filed: https://github.com/aspnet/Logging/issues/688

NinoFloris:
Cool don't mind how it's fixed, saw that you already had some issues open around flowing IncludeScopes config. When filed I hoped this quick fix could still make RTM but alas. 
Thanks for picking it up

PR #211, opened 2017-08-02
Fix __get_remote_file logic

JunTaoLuo:
Similar to https://github.com/aspnet/BuildTools/pull/352. This needs mass update?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -99,17 +99,16 @@ __get_remote_file() {
         return 0
     fi
 
-    failed=false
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -99,17 +99,16 @@ __get_remote_file() {
         return 0
     fi
 
-    failed=false
+    succeeded=false
```
natemcmaster: As long as you're here, can you scope this variable? `local succeeded=false`

Issue #37, opened 2017-04-20
Generate api check baseline for Microsoft.AspNetCore


JunTaoLuo:
Once we release the WebHost static APIs, we should generate an API baseline.

muratg:
Good call! The APIs are new in 2.0.0, so we can do this in 2.0.0 RTM and start to catch breaking changes after that.

muratg:
This will be a "new package" in 2.0.0, so technically we only care about breaking changes after 2.0.0 is done. Moving to 2.1.

PR #213, opened 2017-08-04
Add API baseline for Microsoft.AspNetCore

JunTaoLuo:
https://github.com/aspnet/MetaPackages/issues/37

cc @ryanbrandenburg, apparently you are doing this for all the repos? Should I make this change or wait for yours?


JunTaoLuo:
Got an offline :shipit: from @ryanbrandenburg.

PR #214, opened 2017-08-07
Use custom KoreBuild and dotnet-install if specified

JunTaoLuo:
This allows me to use a custom dotnet-install that skips the installation of the 1.0.5/1.1.2 runtime on certain platforms during the build on ubuntu.16.10.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -857,6 +857,7 @@
       --rm
       -v $(RepositoryRoot):$(DockerRootDirectory)
       -e DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
+      -e KOREBUILD_SKIP_OLD_RUNTIME_INSTALL=true
       -e BuildNumber=$(BuildNumber)
       -e RUNTIMESTORE_TIMESTAMP_INSTALLERNAME=$(RuntimeStoreTimestampInstallerPackageName)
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
Test failures are due to an un-updated M.A.Server.IntegrationTesting package see https://github.com/aspnet/Hosting/pull/1161

natemcmaster:
cc @ryanbrandenburg 

FYI - we're also working on further changes to the KoreBuild bootstrapper to enable docker build. (see https://github.com/aspnet/BuildTools/issues/337). If we finish that before https://github.com/aspnet/BuildTools/issues/347, this work would break this solution. I think it would be better to keep the bootstrappers consistent with other repos, and instead add an env variable such as KOREBUILD_SKIP_LEGACY_DOTNET_INSTALL=1 that simply prevents [this from running](https://github.com/aspnet/BuildTools/blob/dev/files/KoreBuild/scripts/get-dotnet.sh#L77-L79).

JunTaoLuo:
Agreed, will make use of aspnet/BuildTools#360 instead

JunTaoLuo:
Merging now to fix the RuntimeStore_Installer builds.

natemcmaster:
Retroactive :shipit:

PR #220, opened 2017-08-15
Revert Antares test bundle changes

pakrym:
It will go to AzureIntegration


MetaPackages.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26727.1
+VisualStudioVersion = 15.0.26510.0
```
JunTaoLuo: this is an older version?

pakrym: It may be, I just reverted both commits and it was part of them.

PR #225, opened 2017-08-27
Activate WaitForShutdown in sample app

guardrex:
Noticed while working on the hosting topic that u prob meant to unblock these when https://github.com/aspnet/Hosting/pull/1114 merged.

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -22,6 +22,8 @@ namespace SampleApp
             CustomRouter();
 
             CustomApplicationBuilder();
+            
+            CustomMiddlewareDelegate();
 
             StartupClass(args);
         }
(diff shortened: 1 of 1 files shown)


samples/SampleApp/Program.cs
```diff
@@ -22,15 +22,17 @@ public static void Main(string[] args)
...
         }
 
         private static void HelloWorld()
         {
-            using (WebHost.Start(context => context.Response.WriteAsync("Hello, World!")))
+            using (var host = WebHost.Start(context => context.Response.WriteAsync("Hello, World!")))
             {
-                //host.WaitForShutdown(); // TODO: https://github.com/aspnet/Hosting/issues/1022
+                host.WaitForShutdown();
                 Console.WriteLine("Running HelloWorld: Press any key to shutdown and start the next sample...");
```
Tratcher: This needs more cleanup, it prevents the Console.WriteLine, the message is outdated, and is redundant with the ReadKey

guardrex:
The Console.WriteLine is reached if the concept was that the dev issues a Crtl-c in the console. I'll let u cats handle this.

PR #239, opened 2017-09-26
Remove All metapackage and build infrastructure

JunTaoLuo:
- [x] Remove runtime store generation

- [x] Remove LZMA generation

- [x] Remove runtime store installer generation

All of the above will be moved to Universe. Haven't ported the last two components yet but we don't need to wait before removing it from here.


natemcmaster:
Should we (yet-again) change our branch name to rel/2.0.1?

JunTaoLuo:
Yea good point, it should be renamed to 2.0.1. I'll rename it after our patch branches starts building again.

PR #244, opened 2017-10-25
Remove All metapackage and runtime store infrastructure

natemcmaster:
This is being ported to aspnet/Universe. 

It's flaky at the moment because it uses floating package reference versions, so let's pause this until we can fixup aspnet/Universe.


src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
```diff
@@ -1,29 +0,0 @@
-<Project>
```
JunTaoLuo: Either we need to remove this from Universe's artifact.props https://github.com/aspnet/Universe/blob/a01aee7e9c92b64017c71c740668ab688c42a26f/build/artifacts.props#L12 or we need to keep this. I'd imagine this would fail the "coherence" check.

build/repo.targets
```diff
@@ -1,970 +0,0 @@
...
-
-    <ItemGroup>
-      <PackageToTrim Include="runtime.win-arm64.runtime.native.system.data.sqlclient.sni" />
-    </ItemGroup>
-
-    <MSBuild Projects="$(RuntimeStoreReferenceDirectory)\Build.RuntimeStore.References.csproj" Targets="GetPackageDefinitions" >
-      <Output TaskParameter="TargetOutputs" ItemName="_PackageDefinitions" />
-    </MSBuild>
-
-    <RepoTasks.CreateCommonManifest DestinationFilePath="$(ArtifactsDir)$(CommonManifestFileName)" PackageDefinitions="@(_PackageDefinitions)" Packages="@(PackageToTrim)"/>
```
JunTaoLuo: We can remove the repo tasks as well.

src/Microsoft.AspNetCore.All/build/PublishWithAspNetCoreTargetManifest.targets
natemcmaster: I'll update it before I merge this.

build/repo.targets
natemcmaster: Gone.

natemcmaster:
version.xml cleanup is in https://github.com/aspnet/MetaPackages/pull/245

PR #250, opened 2017-12-21
Add kestrel config to CreateDefaultBuilder

Tratcher:
Kestrel now supports config, but it doesn't read any by default. CreateDefaultBuilder will specify the default location of the Kestrel config section. @DamianEdwards, @danroth27 asked if this config section should be namespaced as "Microsoft:Kestrel" or if a top level "Kestrel" section was ok?

I've also updated an old appsettings.json sample with the new config schema.

See https://github.com/aspnet/KestrelHttpServer/pull/2186

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -115,7 +115,7 @@ namespace Microsoft.AspNetCore
         /// </summary>
         /// <remarks>
         ///   The following defaults are applied to the returned <see cref="WebHostBuilder"/>:
-        ///     use Kestrel as the web server,
+        ///     use Kestrel as the web server and configure it using the application's configuration providers,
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -148,7 +148,10 @@ private static IWebHost StartWith(string url, Action<IServiceCollection> configu
         public static IWebHostBuilder CreateDefaultBuilder(string[] args)
         {
             var builder = new WebHostBuilder()
-                .UseKestrel()
+                .UseKestrel((builderContext, options) =>
```
DamianEdwards: Update the remarks above (and related ones on other overloads) to mention that Kestrel will be configured, e.g.:
`///     use Kestrel as the web server and configure it using the application's configuration providers,`

Tratcher: Updated

PR #251, opened 2017-12-22
Force http in tests

Tratcher:
The prior change did fix the invalid cert issue, but then some tests started failing only on net461 with an algorithm conflict. @javiercn you may want to check the default algorithms between a net461 client and a Core server.

Different fix: force these tests to use http.

Feel free to merge, I may be offline.


javiercn:
@muratg I'm already filing

javiercn:
@muratg https://github.com/aspnet/MetaPackages/issues/252

I do believe what @Tratcher means is that there is a conflict between the default algorithms used by HTTP Client when establishing an HTTPS connection to a Kestrel server on the edge exposing an HTTPS endpoint. Maybe @halter73 knows something about it?

javiercn:
Seems like the build is hanging on appveyor. Is there any way to re-trigger it? @ryanbrandenburg

javiercn:
I've pushed an empty commit to the branch to re-trigger the appveyor build

PR #257, opened 2018-01-30
Update WebHost.cs

Matthew-Bonner:
Only add console and debug logging when the environment is development.


Eilon:
@DamianEdwards / @davidfowl / @muratg / @JunTaoLuo - do we plan to take any action on this PR?

Eilon:
My personal view is that we should close it because we want the Console logger to write out in scenarios such as Azure App Services, but I'm no expert.

Also, we generally avoid doing any logic based on the deployment environment. (Not that we *never* do, but it's rare.)

muratg:
Agreed. @Matthew-Bonner thanks for the PR, but like Eilon said, there are scenarios where this is useful. We'll be keeping it in. It's easy to workaround this if it's causing any issues.

PR #261, opened 2018-03-23
Add HostFiltering to the default web host

Tratcher:
Damian didn't want this code in the template so we're putting it in the default builder instead.
https://github.com/aspnet/templating/pull/374

It will be disabled by default, it only lights up if you define AllowedHosts as a semicolon separated list in config.

### DIFF
--- a/samples/SampleApp/appsettings.json
+++ b/samples/SampleApp/appsettings.json
@@ -1,4 +1,5 @@
 {
+  "AllowedHosts":  "example.com;localhost",
   "Kestrel": {
     "EndPoints": {
       "Http": {
--- a/src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
+++ b/src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="$(MicrosoftAspNetCoreDiagnosticsPackageVersion)" PrivateAssets="None" />
+    <PackageReference Include="Microsoft.AspNetCore.HostFiltering" Version="$(MicrosoftAspNetCoreHostFilteringPackageVersion)" PrivateAssets="None" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="$(MicrosoftAspNetCoreHostingPackageVersion)" PrivateAssets="None" />
     <PackageReference Include="Microsoft.AspNetCore.Routing" Version="$(MicrosoftAspNetCoreRoutingPackageVersion)" PrivateAssets="None" />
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -182,6 +182,16 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
                     logging.AddConsole();
                     logging.AddDebug();
                 })
+                .ConfigureServices((hostingContext, services) =>
+                {
+                    var hosts = hostingContext.Configuration["AllowedHosts"]?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
```
JunTaoLuo: Hmm hardcoded constants look suspicious. Why isn't `AllowedHosts` defined as a static string somewhere?

samples/SampleApp/appsettings.json
```diff
@@ -1,4 +1,5 @@
 {
+  "AllowedHosts":  "example.com;localhost",
```
jkotalik: nit extra space.

JunTaoLuo:
@jkotalik tests were modified to check for host filtering as part of their test step.

src/Microsoft.AspNetCore/WebHost.cs
Tratcher: It's only used in this one spot. There are several of those in this method.

src/Microsoft.AspNetCore/Microsoft.AspNetCore.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="$(MicrosoftAspNetCoreDiagnosticsPackageVersion)" PrivateAssets="None" />
+    <PackageReference Include="Microsoft.AspNetCore.HostFiltering" Version="$(MicrosoftAspNetCoreHostFilteringPackageVersion)" PrivateAssets="None" />
```
davidfowl: This makes me kinda sad.

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -182,6 +182,16 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
...
+                    services.AddHostFiltering(options =>
+                    {
+                        options.AllowedHosts = (hosts?.Length > 0 ? hosts : new[] { "*" });
+                    });
+
+                    services.AddTransient<IStartupFilter, HostFilteringStartupFilter>();
```
davidfowl: Can you make sure this middleware is optimized for nooping?

Tratcher: It is. https://github.com/aspnet/BasicMiddleware/blob/2f92fc1d1eaa4033e5b02581da0018dcf8511ea1/src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs#L150

PR #263, opened 2018-04-11
Reload HostFilteringOptions on change

Tratcher:
The second half of https://github.com/aspnet/BasicMiddleware/pull/320. There's no direct dependency between these, they can be checked in independently.

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -5,12 +5,14 @@ using System;
 using System.IO;
 using System.Reflection;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.HostFiltering;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -184,11 +186,20 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
...
+                        if (options.AllowedHosts == null || options.AllowedHosts.Count == 0)
+                        {
+                            // "AllowedHosts": "localhost;127.0.0.1;[::1]"
+                            var hosts = hostingContext.Configuration["AllowedHosts"]?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+                            // Fall back to "*" to disable.
+                            options.AllowedHosts = (hosts?.Length > 0 ? hosts : new[] { "*" });
+                        }
                     });
+                    // Change notification
+                    services.AddSingleton<IOptionsChangeTokenSource<HostFilteringOptions>>(
```
HaoK: If you want to avoid this explicit wireup, you can use the config overload which wires up the change notification for you:

`services.Configure<HostFilteringOptions>(hostingContext.Configuration, _ => { })`

HaoK: Unfortunately there's no overload for PostConfigure that does this.

we could add something to options builder to roll all of this up into a new one call:


```
services.AddOptions<HostFilteringOptions>.PostConfigure(hostingContext.Configuration, options => { 
// your stuff
});
```

Tratcher: A customized config binder?

Tratcher: That would be nice, but would you expect anybody else to use it?

HaoK: This is mostly polish, we already added the OptionsBuilder for 2.1, but we probably have missed some various overloads etc like this.

In general the OptionsBuilder will be the new public surface area, basically similar to what we did with AddAuthentication() returning a builder, the main benefit is that it moves the option type and option name to the first call, and then you can chain Configure/PostConfigure/Bind/whatevers to your heart's content.

PR #268, opened 2018-05-24
 Do not overwrite ContentRoot

### LINKED ISSUE #267: CreateDefaultBuilder overrides ASPNETCORE_CONTENTROOT
Previously reported at https://github.com/aspnet/Hosting/issues/1335.

CreateDefaultBuilder always calls `.UseContentRoot(Directory.GetCurrentDirectory())`, which means it overwrites ASPNETCORE_CONTENTROOT that was read in the WebHostBuilder constructor.

Some recent changes in the functional test infrastructure have started relying on ASPNETCORE_CONTENTROOT (https://github.com/aspnet/Hosting/pull/1429) and now WebHost functional tests are broken. (http://aspnetci/viewLog.html?buildId=474296&tab=buildResultsDiv&buildTypeId=Lite_UniverseTest)

Proposal: Only call UseContentRoot if a value has not already been provided.

Tratcher:
Fixes #267 and http://aspnetci/viewLog.html?buildId=474296&tab=buildResultsDiv&buildTypeId=Lite_UniverseTest

Honor ASPNETCORE_CONTENTROOT, and only set CurrentDirectory if no other value was provided.

I consolidated the config code for clarity.

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -149,12 +149,21 @@ namespace Microsoft.AspNetCore
         /// <returns>The initialized <see cref="IWebHostBuilder"/>.</returns>
         public static IWebHostBuilder CreateDefaultBuilder(string[] args)
         {
-            var builder = new WebHostBuilder()
-                .UseKestrel((builderContext, options) =>
+            var builder = new WebHostBuilder();
+
+            if (string.IsNullOrEmpty(builder.GetSetting(WebHostDefaults.ContentRootKey)))
+            {
+                builder.UseContentRoot(Directory.GetCurrentDirectory());
+            }
+            if (args != null)
+            {
+                builder.UseConfiguration(new ConfigurationBuilder().AddCommandLine(args).Build());
+            }
+
+            builder.UseKestrel((builderContext, options) =>
                 {
                     options.Configure(builderContext.Configuration.GetSection("Kestrel"));
                 })
-                .UseContentRoot(Directory.GetCurrentDirectory())
                 .ConfigureAppConfiguration((hostingContext, config) =>
                 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -149,12 +149,21 @@ private static IWebHost StartWith(string url, Action<IServiceCollection> configu
         /// <returns>The initialized <see cref="IWebHostBuilder"/>.</returns>
         public static IWebHostBuilder CreateDefaultBuilder(string[] args)
         {
-            var builder = new WebHostBuilder()
-                .UseKestrel((builderContext, options) =>
+            var builder = new WebHostBuilder();
+
+            if (string.IsNullOrEmpty(builder.GetSetting(WebHostDefaults.ContentRootKey)))
+            {
+                builder.UseContentRoot(Directory.GetCurrentDirectory());
```
JunTaoLuo: Technically this would be a breaking change so are we okay with taking this @DamianEdwards @davidfowl ? Spoke offline with @glennc and this seems to be okay to take in 2.2.

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -149,12 +149,21 @@ private static IWebHost StartWith(string url, Action<IServiceCollection> configu
...
+            }
+            if (args != null)
```
davidfowl: Why did you move this?

Tratcher: Only for clarity so all of the config would be together, it's not a functional change.

src/Microsoft.AspNetCore/WebHost.cs
Tratcher: I cleared this with @DamianEdwards offline.

PR #274, opened 2018-07-05
Remove doc about unshipped default configuration

poke:
This removes the documentation about the default configuration capabilities in the default web host builder. The feature was originally introduced in 4b18cf52ae3c22c7124fd9cb35ae0253b390b28e and soon unshipped in c920c459537c21285b47b05f033a48363297e3dc. However, the documentation that was added with it never got removed.

Since this is visible on [docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webhost.createdefaultbuilder?view=aspnetcore-2.0), we should probably fix that before more people look for features that aren’t there.

/cc @ctolkien


Tratcher:
Good catch. Can you rebase on release/2.2? Our branching just got overhauled. I retargeted the PR but it has extra commits.

poke:
Oh, I totally forgot about the branching change. Rebased!

PR #278, opened 2018-07-25
Add UseIIS to CreateDefaultBuilder

pakrym:
https://github.com/aspnet/MetaPackages/issues/279

cc @shirhatti

### DIFF
--- a/src/Microsoft.AspNetCore/WebHost.cs
+++ b/src/Microsoft.AspNetCore/WebHost.cs
@@ -210,6 +210,7 @@ namespace Microsoft.AspNetCore
 
                     services.AddTransient<IStartupFilter, HostFilteringStartupFilter>();
                 })
+                .UseIIS()
                 .UseIISIntegration()
                 .UseDefaultServiceProvider((context, options) =>
                 {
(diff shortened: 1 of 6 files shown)


test/Microsoft.AspNetCore.FunctionalTests/WebHostFunctionalTests.cs
```diff
@@ -165,6 +166,42 @@ public void LoggingConfigurationSectionPassedToLoggerByDefault()
...
+
+            using (var deployer = IISApplicationDeployerFactory.Create(deploymentParameters, LoggerFactory))
+            {
+                var deploymentResult = await deployer.DeployAsync();
+                var response = await RetryHelper.RetryRequest(() => deploymentResult.HttpClient.GetAsync(string.Empty), Logger, deploymentResult.HostShutdownToken, retryCount: 5);
+
+                var responseText = await response.Content.ReadAsStringAsync();
+                try
+                {
+                    // Assert server is Kestrel
```
jkotalik: nit: comment is wrong.

src/Microsoft.AspNetCore/WebHost.cs
```diff
@@ -210,6 +210,12 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
 
                     services.AddTransient<IStartupFilter, HostFilteringStartupFilter>();
                 })
+#if NETCOREAPP2_2
```
Tratcher: You don't actually need this conditional, Server.IIS targets netstandard2.0. If Server.IIS has specific TFM requirements they should be handled internally.

pakrym:
:up::date:

PR #295, opened 2018-10-30
Don't use the default 5000 port

JunTaoLuo:
Addresses: https://github.com/aspnet/MetaPackages/issues/294 by using dynamic ports instead.


test/Microsoft.AspNetCore.FunctionalTests/WebHostFunctionalTests.cs
```diff
@@ -4,31 +4,20 @@
...
 {
     public class WebHostFunctionalTests : LoggedTest
     {
-        private readonly string _testSitesPath;
-
-        public WebHostFunctionalTests(ITestOutputHelper output) : base(output)
-        {
-            _testSitesPath = GetTestSitesPath();
-        }
-
```
JunTaoLuo: Note that there are two other tests `CreateDefaultBuilder_InitializeWithDefaults` and `CreateDefaultBuilderOfT_InitializeWithDefaults` that uses a hardcoded port 5002 to ensure the default app settings are loaded. I'll see if I can use another setting instead since hard coding ports is a bad pattern.

JunTaoLuo:
Updated the other tests.

test/TestSites/CreateDefaultBuilderApp/Program.cs
```diff
@@ -15,24 +14,28 @@ public class Program
...
-                .ConfigureServices((context, services) =>
+                .ConfigureServices((context, services) => responseMessage = responseMessage ?? GetResponseMessage(context))
+                .ConfigureKestrel(options =>
                 {
-                    responseMessage = GetResponseMessage(context, services);
-                })
-                .Configure(app =>
-                {
-                    app.Run(context =>
+                    if (responseMessage == null
```
Tratcher: change this to use the [Endpoint](https://github.com/aspnet/KestrelHttpServer/blob/e958d8258445c2cda73c244d586f06ca6753be4e/samples/SampleApp/Startup.cs#L147-L151) API to make sure the endpoints actually get loaded:
options.Endpoint("HTTP", opt =>
                        {
                            opt.ConfigSection["KestrelEndPointSettingName"] ...
                        })

PR #298, opened 2018-11-16
Added a method to configure the web host defaults

davidfowl:
- Added `ConfigureWebHostDefaults` to configure the web specific defaults on the generic host builder.

An alternative approach is to add a UseWebHostDefaults that just configures these things so we don't have 2 methods:

Option 1

```C#
public static IHostBuilder CreateHostBuilder(string[] args) =>
    Host.CreateDefaultBuilder(args)
        .ConfigureWebHostDefaults(builder =>
        {
            builder.UseStartup<Startup>();
        });
```

Option 2

```C#
public static IHostBuilder CreateHostBuilder(string[] args) =>
    Host.CreateDefaultBuilder(args)
        .UseWebHostDefaults()
        .ConfigureWebHost(builder =>
        {
            builder.UseStartup<Startup>();
        });
```

PS: This doesn't do the defaults that aren't specific to the web stuff. We still need to figure out where that goes.


Tratcher:
Update https://github.com/aspnet/MetaPackages/blob/master/samples/SampleApp/Program.cs

Tratcher (approved):
The division looks right. We can't update templates until we have both parts.

Tratcher:
As for `Host.CreateDefaultBuilder(args)`, all its dependencies are in Extensions so how about adding a new package in Hosting called `Microsoft.Extensions.Hosting.Defaults` that pulls it together?

